PR #2, opened 2016-02-04
Update Makefile

elfprince13:
Default path for CEDev doesn't make any sense given the location of bin compared to this makefile.


elfprince13:
Note, this update is unrelated to #6

MateoConLechuga:
Git hell sucks

elfprince13:
Something went horribly wrong. You're missing the line that was supposed to be updated.

PR #19, opened 2016-02-24
Improved demo0 "purity", moved template-like stuff to a template dir

KermMartian:
A few initial improvements to the demo directory:
- Made demo0 less of a template and more of a demo. I still need to debug
  why it crashes on quit.
- Simplified the template-ish parts of demo0 into a new template directory.


MateoConLechuga:
Nice, I think I know why it is crashing. I'll have to run some quick tests though. Thanks! :)

PR #28, opened 2016-10-17
Optimized _TransparentPlot_ASM for speed*

Runer112:
Unrolled each of the opaque and transparent braches into four iterations, so stretches of the same pixel opaqueness/transparency mostly fall through the jumps and execute faster.

Before (12 bytes):
 * opaque->opaque: 41 cc
 * opaque->trans: 36 cc
 * trans->opaque: 46 cc
 * trans->trans: 41 cc

After (54 bytes; lists cycles after 1/2/3/4 mod 4 iterations in a row of previous pixel state):
 * opaque->opaque: 36/36/36/41 cc 
 * opaque->trans:  41/41/41/36 cc
 * trans->opaque: 38/38/38/46 cc
 * trans->trans: 33/33/33/38 cc

Starts on the fast transparent path based on the assumption that the majority of transparent sprite pixel rows start with a transparent pixel, as transparency is often used to draw "cutout" images.

### DIFF
--- a/CEdev/lib/src/libraries/graphics_src/graphx/v2/graphics_lib.asm
+++ b/CEdev/lib/src/libraries/graphics_src/graphx/v2/graphics_lib.asm
@@ -1796,16 +1796,50 @@ ClipSprTransNextAmt =$+1
 	pop	ix
 	ret
 
-_:	ldi
+_TransparentPlot_ASM_Opaque:
+	ldi
 	ret	po
-_TransparentPlot_ASM:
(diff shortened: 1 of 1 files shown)


Runer112:
Untested; would appreciate if someone could at least take a second look and check that it looks reasonable.

MateoConLechuga:
Thanks! :)

Runer112:
For the record: @MateoConLechuga tested this with an example program using transparent sprites and said that it worked.

PR #31, opened 2016-10-22
SwapDraw improvement

TheMachine02:
*should* be smaller and a bit faster, but I also may broke everything...


MateoConLechuga:
This breaks everything :p

TheMachine02:
Arg, was pretty sure I overlooked something >_>

PR #32, opened 2016-10-23
Optimisation of _Rectangle_NoClip

grosged:
I wrote it , with optimisation/priority for common cases in mind ;)


grosged:
Finally,  very littlte little gain....So ...useless

PR #34, opened 2016-10-25
suggest a faster FillCircle_NoClip routine

grosged:
Seems to be 10% faster


MateoConLechuga:
This routine destroys ix...

PR #35, opened 2016-10-30
suggest a faster version  of  ScaledSprite_NoClip

grosged:
appears to be 15% faster


MateoConLechuga:
Technically the null checks on the width and height can be removed; I just assume the user knows what they are doing :P

grosged:
Plus, there was a mistake : read width scale at height scale offset !

MateoConLechuga:
If you would be so kind to move this routine to its new location that would be great :)

grosged:
I will come back

PR #36, opened 2016-10-31
suggest a faster version of Sprite_NoClip

grosged:
about 5% faster


MateoConLechuga:
Fix the final ret and the tab spacing :)

grosged:
I will come back

PR #37, opened 2016-11-03
optimisation of imuls|imulu (8,5% faster)

grosged:


adriweb:
So hum, Mateo reported that it actually crashes - Why hasn't this been tested [enough] ? :/
I assume grosged was able to come up with benchmarks by having it run correctly, though, so are there any specific conditions for the crash?
In any case, such a function should pass standard tests with "standard" values as well as edge cases - and it should at least behave exactly like the old one (unless it fixes a bug of course)

For these kinds of functions, it would really be great to be able to unit-test them automatically, since there's not even a TI ROM needed, actually. But setting that up might be a bit hard/challenging (but it's a good challenge :P)

PR #40, opened 2016-11-07
optimised _PrintUInt & _FillScreen

grosged:
Slight optimisation of _PrintUInt,
Propose to include a faster fillscreen when color is null


CEdev/lib/src/graphics/graphx/v2/graphics_lib.asm
```diff
@@ -1983,13 +1991,13 @@ _ClipDraw_ASM:
 ;  DE : New X coordinate
 ;  NC : If offscreen
 	ld	ix,6                        ; get pointer to arguments
+	ld	iy,ix-6
```
adriweb: This doesn't exist :D

MateoConLechuga: I'm just going to merge anyway and fix myself

PR #41, opened 2016-11-18
Update graphics_lib.asm

PeterTillema:
Small optimizations ;)


MateoConLechuga:
This doesn't actually work :p

PR #62, opened 2017-04-29
Reset iy for token hook

PeterTillema:
_Get_Tok_Strng calls the token hook, which needs bit 0, (iy+$35) or so, and if iy is random, it might happen that the OS things the hook is set, while it isn't, and then it resets the bit.


MattWaltz:
Thanks :)

PR #72, opened 2017-06-26
Optimize gfx_FillScreen and gfx_ZeroScreen

Runer112:
Sanity check, please.


MattWaltz:
spaces not tabs plz :)

Runer112:
We use tabs in instructions. I don't see why not to use them for indenting comments. It's a hell of a lot easier to make comments this way.

PR #74, opened 2017-06-28
Improvement/rand

Runer112:


Runer112:
Do not merge yet. Waiting to hear back from the creator of XSadd if there are shift parameters that can be used that result in more optimal code. Feedback on its current state is still welcome, though.

PR #75, opened 2017-06-28
Calculate green LSB in default palette colors

Runer112:
Can someone confirm that this is both correct and desired?

Also, this is a lazy way of running all the autotests to see if any need to be updated.


MattWaltz:
Tested and I actually think it looks a lot better than before. Awesome :)

PR #83, opened 2017-07-31
Optimize few things

PeterTillema:


src/graphx/graphx.asm
```diff
@@ -829,22 +829,21 @@ _SetDraw:
 ;  arg0: buffer or screen
 ; Returns:
 ;  None
+	pop	de
 	pop	hl
```
runer112: This should be `ex (sp),hl` instead of `pop hl`. It is the caller's responsibility to deallocate the argument stack space, not the callee's. Callees must leave the stack space in place (although you can overwrite the data, as `ex (sp),hl` does).

adriweb:
_(Note that tests won't pass, because for travis to be able to fetch the ROM, ie. have access to its secure env var thingy, it needs to come from this repo, not a fork ; so PRs coming from here but another branch are fine)_

PR #87, opened 2017-08-02
Few opts

PeterTillema:


src/graphx/graphx.asm
```diff
@@ -460,7 +459,6 @@ _GetPixel:
 	inc	hl
 	inc	hl
 	inc	hl                          ; move to next argument
-	ld	de,0
```
runer112: This removal is not safe. Although _PixelPtr_ASM is documented as the y input being only 8-bit, the full 24-bit value is used for clipping.

src/graphx/graphx.asm
```diff
@@ -481,7 +479,6 @@ _SetPixel:
...
```
runer112: Ditto.

src/graphx/graphx.asm
```diff
@@ -648,12 +645,12 @@ _Rectangle_NoClip:
 ;  arg3 : Height
 ; Returns:
 ;  None
-	ld	iy,0
+	ld	iy,-1
```
runer112: Why..?

src/graphx/graphx.asm
```diff
@@ -679,36 +676,37 @@ _HorizLine:
 ;  arg2 : Length
 ; Returns:
 ;  None
-	ld	iy,0
+	ld	iy,-1
```
runer112: Why..?

src/graphx/graphx.asm
```diff
@@ -5542,8 +5540,8 @@ _PixelPtr_ASM:
 	ld	hl,-lcdWidth
 	add	hl,bc
 	ret	c
-	ld	hl,-lcdHeight
-	add	hl,de
+	ld	a,256-lcdHeight
```
runer112: I believe that some usages expect clipping on the full 24-bit value.

src/graphx/graphx.asm
```diff
@@ -5542,8 +5540,8 @@ _PixelPtr_ASM:
...
+	add	a,e
```
runer112: _Circle calls _SetPixel_ASM, which calls this (_PixelPtr_ASM). Although this routine is documented as y only being 8-bit, _Circle needs it to be 24-bit for proper clipping.

mateoconlechuga:
This is because clipping an 8-bit value when it is already out of range of an 8-bit signed number is not doable.

PR #94, opened 2017-09-23
std/shared: dirty fix for linkage issues with __print_send

tqbunny:


mateoconlechuga:
CI doesn't work entirely for external PRs yet; but this seems to be good :)

PR #100, opened 2017-12-01
Fix more typos

PeterTillema:


adriweb:
Next time, please fixup all your commits so there's just one ;)
I'll rebase-merge this so it'll be the same, but others might forget about that feature.

PR #115, opened 2018-02-10
slight optimisations

grosged:

PR #116, opened 2018-02-10
(grosged) slight optimisations

adriweb:
imported PR from https://github.com/CE-Programming/toolchain/pull/115 so that travis can run the tests

### DIFF
--- a/src/graphx/graphx.asm
+++ b/src/graphx/graphx.asm
@@ -202,14 +202,16 @@ gfx_AllocSprite:
 ;  arg2 : pointer to malloc routine
 ; Returns:
 ;  Pointer to allocated sprite, first byte width, second height
-	ld	bc,3
-	push	bc
-	pop	hl
+	ld	hl,3
 	add	hl,sp
 	ld	e,(hl)			; e = width
-	add	hl,bc
+	inc	hl
+	inc	hl
+	inc	hl
 	ld	d,(hl)			; d = height
-	add	hl,bc
+	inc	hl
+	inc	hl
+	inc	hl
 	ld	hl,(hl)			; hl = malloc
 	push	de
 	mlt	de			; de = width * height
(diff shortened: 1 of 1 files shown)


src/graphx/graphx.asm
```diff
@@ -523,13 +525,15 @@ gfx_GetPixel:
 	inc	hl
 	inc	hl
 	inc	hl			; move to next argument
-	ld	de,0
 	ld	e,(hl)			; e = y coordinate
-	call	_PixelPtr
```
adriweb: I believe this routine does some checks first, here is it safe to get rid of that?

grosged: My suggestion : not verifying x,y coordinates as the result is just a value of one pixel on screen. Until now, If coord outside of screen, what's the result ?

grosged: when coordinates are definitely outside, then what's the returned value for color , please ?

grosged: to check or not to check ?..That's the question :)

src/graphx/graphx.asm
```diff
@@ -544,8 +548,7 @@ gfx_SetPixel:
 	inc	hl
 	inc	hl
 	inc	hl			; move to next argument
-	ld	de,0
-	ld	e,(hl)			; e = y coordinate
+	ld	de,(hl)			; e = y coordinate
```
runer112: _PixelPtr checks the full 24-bit y value in DE, but the y value is only passed as an 8-bit value to gfx_SetPixel. The upper 16 bits of DE will be undefined and possibly nonzero here, in which case I believe _PixelPtr would erroneously reject it as offscreen even if the 8-bit value is actually on screen.

grosged: I see. Please reject it. If I remember well, I suggested that twice.

src/graphx/graphx.asm
runer112: It seems a little sketchy that this could read from an arbitrary memory address, but jacobly seems to think that should be okay.

runer112:
All opts I didn't comment on look good.

PR #117, opened 2018-02-10
_MultiplyHLDE a little bit faster

grosged:


mateoconlechuga:
You can't do that in fasmg!

src/graphx/graphx.asm
```diff
@@ -5888,7 +5890,7 @@ _MultiplyHLBC:
 	push	iy
 	push	hl
 	push	bc
-	push	hl
+DEused:	push	hl
 	ld	iy,0
 	ld	d,l
 	ld	e,b
```
runer112: B is used here, so this change isn't actually valid.

PR #119, opened 2018-02-10
Update graphx.asm

grosged:
optimised gfx_RotateSpriteHalf

But I'm not sure about that point when ending the routine:
 "arg1 : Pointer to sprite struct output"
(I can't certify this as "returned value" , please confirm)

### DIFF
--- a/src/graphx/graphx.asm
+++ b/src/graphx/graphx.asm
@@ -4164,10 +4164,13 @@ gfx_RotateSpriteHalf:
 	mlt	bc
 	add	hl,bc
 	lea	de,iy
-ploop:	inc	de
+	push	de
+.loop:
+	inc	de
 	inc	de
 	ldd
-	jp	pe,ploop
+	jp	pe,.loop
+	pop	hl
 	ret
 
 ;-------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


src/graphx/graphx.asm
```diff
@@ -4156,29 +4156,18 @@ gfx_RotateSpriteHalf:
...
+	ld	(iy+0),bc
 	mlt	bc
-	lea	hl,ix+1
 	add	hl,bc
-	push	ix
-.loop:
-	ld	a,(de)
-	ld	(hl),a
+	lea	de,iy
+ploop:	inc	de
```
runer112: Why was this label changed? Local labels are preferred inside of routines like this, and it's also preferred for labels to exist on their own line.

grosged: ah, I found the "." weird (actually, I realised later fasmg requires it ).
Please revert to ".loop" ;)

grosged: @mateoconlechuga 
Thanks to confirm my doubt ;)
then, I suggest to insert between lines 4166-4167 a "push de" and between lines 4170-4171 a "pop hl" . I think it will be okay :)

PR #123, opened 2018-03-05
_PixelPtr no more sub-routined

grosged:
As subroutine _PixelPtr & included _PixelPtrNoChecks were actually called only once, I suggest no more call for each.


adriweb:
Lol the autotester segfaults now apparently. 
We’ll have to fix that. 

Good job creating the PR in a branch though :)

runer112:
380d3d0 is not a valid optimization. `sp` should not be altered unless you're in a context where interrupts should not occur, and they can definitely occur in this context.

PR #142, opened 2018-06-11
Fix _GetC

PeterTillema:
If filesize = 0 and offset = 0, _GetC should return EOF, and not the next byte in memory.


runer112:
Looks good, but may as well let CI finish first to be sure.

runer112:
Right, CI doesn't work because of silly security reasons. The change is definitely good though, so I'll just accept it anyway.

PR #155, opened 2018-07-25
Properly overwrite the variable type

PeterTillema:
ti_DetectAny( didn't overwrite the variable type, it only writes them to the stack entry, which is wrong obviously. This PR fetches the address of the type var and writes the type of the variable to this address.


adriweb:
@PeterTillema told me this fixes the bug, and so the failing test (see #156) should now pass.

PR #157, opened 2018-07-31
Make LibLoad itself into a library

runer112:
This is paired with the branch of the same name in the libload repo. That PR is [here](https://github.com/CE-Programming/libload/pull/2), and it explains the motivations behind this. This PR should provide some regression testing.


src/startup/libheader.src
```diff
@@ -40,6 +40,7 @@ __notfound:
 
 	segment	libs
 __relocationstart:		; relocation data will be placed here
+	db $C0, "LibLoad", 0, 30 ; hardcode dependency on LibLoad v3.0, FIXME HACK
```
runer112: Any ideas for how to add this dependency in a more proper manner? At least, one that pulls the appropriate version number from libload's source?

runer112:
Although there is no automatic testing, I have manually tested the new combinations of new/old LibLoad and new/old program using LibLoad, and they all work as expected.

src/startup/libheader.src
mateoconlechuga: A fasmg macro to include a file in a virtual space and then pull the symbol or something would do the trick

jacobly0:
- [ ] Update fasmg-ez80 submodule.
- [ ] Convert linker options (`-i "sources '<source>" -i "deps '<dep>'"` → `-i "srcs '<source>' if 1, '<dep>'"`).
- [ ] Add libload as a dependency (`-i "libs 'libload.asm' if libs.length"`)
- [ ] Not required for this PR, but install and use libraries' *.lib files instead their *.asm sources.

src/startup/libheader.src
mateoconlechuga: Is there any good way to resolve this now?

mateoconlechuga:
There's a command line overflow that breaks appveyor, but the goal of this pr is complete. So, merge time.

PR #167, opened 2018-09-15
Fix gfx_Begin to set the right JR offset at larger text scale

PeterTillema:
Currently, if you call `gfx_Begin()` twice, with a `gfx_SetTextScale(1,1)` in between, everything after that second `gfx_Begin()` is not scaled, because the jr offset is overwritten to be 0.

### DIFF
--- a/src/graphx/graphx.asm
+++ b/src/graphx/graphx.asm
@@ -240,7 +240,7 @@ gfx_Begin:
 ; Returns:
 ;  None
 	ld	hl,_LargeFontJump
-	ld	a,(hl)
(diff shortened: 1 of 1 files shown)


runer112:
My comment on the related ICE issue (https://github.com/PeterTillema/ICE/issues/97):

You cannot blame the toolchain. Calling `gfx_Begin()` (aka `det(0)`) twice without calling `gfx_End()` in between is incorrect usage. Any number of bugs may occur because this is unsupported.

Also, even if you do end and then restart graphx, I'd say that whether or not any settings carry over is undefined. Users should re-initialize any settings they need.

PR #172, opened 2018-10-24
Improved MMIO access

runer112:
Fix/add MMIO defines, add some "atomic" 32-bit load functions, and add a stopwatch example.

One main thing that any reviewers should check is that I didn't mess up the changes to the MMIO defines.


examples/stopwatch/src/main.c
```diff
@@ -0,0 +1,60 @@
...
+    /* Float format for printf may be unimplemented, so go through an OS real */
+    real_t elapsed_real;
+
+    /* Max stopwatch value is (2^32 - 1) / 32768 = 131072.00, */
+    /* so create a buffer with room for 9 characters plus a null terminator */
+    char str[10];
+
+    /* If the elapsed time is small enough that the OS would print it using */
+    /* scientific notation, force it down to zero before conversion */
+    elapsed_real = os_FloatToReal(elapsed <= 0.001f ? 0 : elapsed);
```
adriweb:
```suggestion
    elapsed_real = os_FloatToReal(elapsed <= 0.001f ? 0.0f : elapsed);
```

PR #182, opened 2019-01-14
Allow editing the linker_script

PeterTillema:


adriweb:
BTW, target the devel branch.
(and fix the missing parens, apparently :P)

PR #185, opened 2019-03-12
FontLibC

drdnar:
Adds FontLibC to development branch. Should be usably stable and match the style guide, though I comment more than you :p


adriweb:
I'll add an autotest soon.

src/fontlibc/fontlibc.h
```diff
@@ -0,0 +1,437 @@
...
+void fontlib_DrawString(char* str);
+
+/**
+ * Draws a string, up to a maximum number of characters
+ * THIS IS NOT REENTRANT (though if you need that, you're probably not using C)
+ * @param str Pointer to string
+ * @param max_characters Maximum number of characters to attempt to print, may
+ * return early if some other condition requires returning
+ */
+void fontlib_DrawStringL(char* str, int24_t max_characters);
```
adriweb: Shouldn't that be `const char* str`?

src/fontlibc/fontlibc.h
```diff
@@ -0,0 +1,437 @@
...
+ * @param glyph Codepoint
+ */
+void fontlib_DrawGlyph(uint8_t glyph);
+
+/**
+ * Draws a string
+ * THIS IS NOT REENTRANT (though if you need that, you're probably not using C)
+ * @param str Pointer to string
+ */
+void fontlib_DrawString(char* str);
```
adriweb: Shouldn't that be `const char* str`?

src/fontlibc/fontlibc.h
```diff
@@ -0,0 +1,437 @@
...
+uint8_t fontlib_GetGlyphWidth(char codepoint);
+
+/**
+ * Returns the width of a string printed in the current font.
+ * Stops processing when it encounters ANY control code or a codepoint not in
+ * the current font.
+ * @param str Pointer to string
+ * @return Width of string
+ */
+uint24_t fontlib_GetStringWidth(char* str);
```
adriweb: Shouldn't that be `const char* str`?

src/fontlibc/fontlibc.h
```diff
@@ -0,0 +1,437 @@
...
+/**
+ * Returns the width of a string printed in the current font.
+ * Stops processing when it encounters ANY control code or a codepoint not in
+ * the current font, or when max_characters have been processed.
+ * @param str Pointer to string
+ * @param max_characters Maximum number of characters to process
+ * (or 0 if not needed)
+ * @return Width of string
+ */
+uint24_t fontlib_GetStringWidthL(char* str, int24_t max_characters);
```
adriweb: Shouldn't that be `const char* str`?

src/fontlibc/fontlibc.h
```diff
@@ -0,0 +1,437 @@
...
+#include <stdint.h>
+#include <stdbool.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define FONTLIB_ENABLE_AUTO_WRAP 1
+#define FONTLIB_AUTO_CLEAR_TO_EOL 2
+#define FONTLIB_PRECLEAR_NEWLINE 4
```
runer112: These should ideally be enum values.

src/fontlibc/fontlibc.h
```diff
@@ -0,0 +1,437 @@
...
+	SERIF = 0x01,
+	/* If both are set, then assume there's no difference between oblique
+	 * and italic styles. */
+	OBLIQUE = 0x02,
+	ITALIC = 0x04,
+	/* Chances are you're not using this library for monospaced fonts.
+	 * But if you are, you'll still have to provide a widths table where
+	 * every byte is the same. */
+	MONOSPACED = 0x08
+} font_styles_t;
```
runer112:
```suggestion
} fontlib_styles_t;
```

Following the pattern of other libraries (which is good for scoping), all definitions (not just functions) should share a name prefix.

src/fontlibc/fontlibc.h
```diff
@@ -0,0 +1,437 @@
...
+	intptr_t font_family_name;
+	intptr_t font_author;
+	/* NOTA BENE: TYPEFACES AND BITMAPPED FONTS CANNOT BE COPYRIGHTED UNDER U.S. LAW!
+	 * This field is therefore referred to as a pseudocopyright.  HOWEVER,
+	 * it IS is applicable in other jusrisdictions, such as Germany. */
+	intptr_t font_pseudocopyright;
+	intptr_t font_description;
+	intptr_t font_version;
+	intptr_t font_codepage;
+} font_metadata;
```
runer112:
```suggestion
} fontlib_metadata_t;
```

src/fontlibc/fontlibc.h
```diff
@@ -0,0 +1,437 @@
...
+	   0xC0: bold*/
+	uint8_t weight;
+	/* Specifies the style of the font.  See enum font_styles */
+	uint8_t style;
+	/* For layout, allows aligning text of differing fonts vertically.
+	   These count pixels going down, i.e. 0 means the top of the glyph. */
+	uint8_t cap_height;
+	uint8_t x_height;
+	uint8_t baseline_height;
+} raw_font;
```
runer112:
```suggestion
} fontlib_font_t;
```

src/fontlibc/fontlibc.h
```diff
@@ -0,0 +1,437 @@
...
+
+typedef struct {
+	char header[8]; /* "FONTPACK" */
+	/* Frankly, if you have more than 127 fonts in a pack, you have a
+	   problem. */
+	uint8_t fontCount;
+	/* Offset from first byte of header */
+	intptr_t metadata;
+	intptr_t font_list;
+} font_pack;
```
runer112:
```suggestion
} fontlib_pack_t;
```

src/fontlibc/fontlibc.h
```diff
@@ -0,0 +1,437 @@
...
+void fontlib_SetWindowFullScreen(void);
+
+/**
+ * Sets the bounds of the window all text will appear in
+ * @param x_min X coord base
+ * @param y_min Y coord base
+ * @param width Width
+ * @param height Height
+ */
+void fontlib_SetWindow(int x_min, uint8_t y_min, int width, uint8_t height);
```
runer112: I'd suggest adding a `fontlib_window_t` struct that encapsulates these four values and using that instead. If you match the layout in the library, then this functions becomes rather trivial in implementation. `fontlib_GetWindow` would be rather trivial as well, and I'd recommend resurrecting that.

runer112: And I'm not entirely sure how ZDS's codegen works when constructing temporary structs, but having the parameter be a `const fontlib_window_t *` might be more optimal.

runer112: Oh, and it's a bit weird having only a bulk setter while having individual getters. Unless there's a good case to be made for the individual getters, I'd suggest removing those (once `fontlib_GetWindow` has been resurrected).

runer112: _(So many comments...)_ If you're not planning to support clipping, `x_min` and `width` should be of type `uint24_t`. If you are, then `y_min` and `height` should be of type `int`.

src/fontlibc/fontlibc.h
```diff
@@ -0,0 +1,437 @@
...
+	OBLIQUE = 0x02,
+	ITALIC = 0x04,
+	/* Chances are you're not using this library for monospaced fonts.
+	 * But if you are, you'll still have to provide a widths table where
+	 * every byte is the same. */
+	MONOSPACED = 0x08
+} font_styles_t;
+
+typedef struct {
+	/* These are standard C-strings.  These may be NULL. */
```
runer112: Then type them as such!

src/fontlibc/fontlibc.h
```diff
@@ -0,0 +1,437 @@
...
+	uint8_t height;
+	/* Total number of glyphs provided. */
+	uint8_t total_glyphs;
+	/* Number of first glyph.  If you have no codepoints below 32, for
+	   example, you can omit the first 32 bitmaps. */
+	uint8_t first_glyph;
+	/* Offset/pointer to glyph widths table.
+	 * This is an OFFSET from the fontVersion member in data format.
+	 * However, it is 24-bits long because it becomes a real pointer upon loading. */
+	intptr_t widths_table;
```
runer112: Use `void *` for opaque pointer types.

runer112: (There are multiple of these in the file, I won't comment them all)

src/fontlibc/fontlibc.h
```diff
@@ -0,0 +1,437 @@
...
+	uint8_t baseline_height;
+} raw_font;
+
+typedef struct {
+	char header[8]; /* "FONTPACK" */
+	/* Frankly, if you have more than 127 fonts in a pack, you have a
+	   problem. */
+	uint8_t fontCount;
+	/* Offset from first byte of header */
+	intptr_t metadata;
```
runer112: Assuming this is a pointer to an array of pointers,

```suggestion
	fontlib_metadata_t **metadata;
```

src/fontlibc/fontlibc.h
```diff
@@ -0,0 +1,437 @@
...
+	intptr_t font_list;
```
runer112: Assuming this is a pointer to an array of pointers,

```suggestion
	fontlib_font_t **font_list;
```

src/fontlibc/fontlibc.h
```diff
@@ -0,0 +1,437 @@
...
+
+/**
+ * Sets the current font
+ * @param font_data Pointer to font data
+ * @param flags Information about how to process the font (unused)
+ * @return Returns false if the font seems invalid for any reason
+ * WARNING: If false is returned, no valid font is currently loaded and trying
+ * to print will print garbage!
+ */
+bool fontlib_SetFont(uint8_t* font_data, int flags);
```
runer112:
```suggestion
bool fontlib_SetFont(const fontlib_font_t *font, unsigned flags);
```

src/fontlibc/fontlibc.h
runer112: And `max_characters` might make more sense as `unsigned`.

src/fontlibc/fontlibc.h
```diff
@@ -0,0 +1,437 @@
...
+ */
+char* fontlib_GetLastCharacterRead(void);
+
+/**
+ * Returns 0 if DrawStringL or GetStringWidthL returned because max_characters
+ * were processed.  Returns a non-zero number otherwise.
+ * @return Either zero, or a non-zero number depending on the reason DrawStringL
+ * or GetStringWidthL returned.
+ */
+int24_t fontlib_GetCharactersRemaining(void);
```
runer112: Unless there's a reason why this could be negative,

```suggestion
unsigned fontlib_GetCharactersRemaining(void);
```

src/fontlibc/fontlibc.h
runer112: https://github.com/CE-Programming/toolchain/pull/185/files#r264843705

src/fontlibc/fontlibc.h
mateoconlechuga: const char *str. Anyone who ties the * to the type and not the variable should be stabbed.

examples/fontlib_example/src/main.c
```diff
@@ -0,0 +1,75 @@
...
+    /* We're just using some simple math to find the center of the text window.
+    Then we find the center of the text to be displayed and use math to make
+    sure it ends up in the center of the window. */
+    fontlib_SetCursorPosition(fontlib_GetWindowWidth() / 2 + fontlib_GetWindowXMin() - (fontlib_GetStringWidth(string) / 2), fontlib_GetCursorY());
+    fontlib_DrawString(string);
+}
+
+/* Main Function */
+void main(void) {
+    char* title = " FONTLIBC ";
```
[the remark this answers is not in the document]
drdnar: I dislike that style because it makes it look like pointers aren't separate type, but whatever, fine.

src/fontlibc/fontlibc.h
drdnar: Yes, it can be negative. PrintString/GetStringWidth internally call PrintStringL/GetStringWidthL but set the maximum length to -1. In that case, the value returned is negative how-many-characters-were-processed.

adriweb:
IMHO, char* is much better. Or with a space between those as a last resort. But tied to the var name? Gross!

src/fontlibc/fontlibc.h
drdnar: Internally, it stores an XMax and YMax, not width and height. The former is more useful for bounds checking, the latter seems more useful for actual layout. I think the individual getters are probably faster to call. I didn't add individual setters because it seems unlikely to be frequently useful, but it at least it wouldn't be hard to add.

runer112: You can still have the public struct use width and height and internally store it as xmax and ymax. The implementation of the bulk getter/setter is no longer a simple copy, but it's still simpler (or at least no more complex) than it is currently. And I'd say that both are simpler from the user's perspective.

Regarding individual getters/setters: neither seems particularly useful to me. If anything the getters seem even less useful than the setters. After all, the programmer set the bounds; they should already know what they are.

src/fontlibc/fontlibc.h
runer112: Neither `fontlib_DrawStringL` nor `fontlib_GetStringWidthL` documents the functionality of accepting a negative `max_characters` or returning a negative value.

runer112: Okay, I see how it works now. But not only is this negative return value case not documented, it's also off by one. You get `-chars_processed - 1`.

Rather than trying to make this fit the narrative of being defined behavior, I'd simply suggest stating that the return value of this function is undefined if the last call to either of these two functions specified no maximum length.

src/fontlibc/fontlibc.h
drdnar: I thought about a struct, but internally they're not packed. I think there's even some code in there that depends on being able to write zeros to subsequent bytes for the Y values.

runer112: That's fine. Due to the width/height and xmax/ymax mismatch, you already need a different underlying representation and will have to keep basically the same underlying implementation. But bundling it all into one struct for bulk operations makes for a better API to the programmer, IMO. (Especially for the bulk getter, which you removed because it was so cumbersome)

src/fontlibc/fontlibc.h
drdnar: My intent was that a font pack would contain related fonts (i.e. a font family in various sizes and styles), and therefore only have one set of metadata.

runer112: I have edited my original suggestion accordingly.

But in that case, where there's just one metadata per font pack, why is this a pointer at all? The metadata struct could just be embedded directly.

src/fontlibc/fontlibc.h
drdnar: I don't want to make it official, so it's not documented. But the idea was that you could also look at how many characters were printed. However, GetLastCharacterRead would be a better choice for figuring that out.

src/fontlibc/fontlibc.asm
```diff
@@ -0,0 +1,1533 @@
...
+fontStruct.version := 0
+fontStruct.height := fontStruct.version + 1
+fontStruct.totalGlyphs := fontStruct.height + 1
+fontStruct.firstGlyph := fontStruct.totalGlyphs + 1
+fontStruct.widthsTablePtr := fontStruct.firstGlyph + 1
+fontStruct.bitmapsTablePtr := fontStruct.widthsTablePtr + 3
+fontStruct.italicSpaceAdjust := fontStruct.bitmapsTablePtr + 3
+fontStruct.spaceAbove := fontStruct.italicSpaceAdjust + 1
+fontStruct.spaceBelow := fontStruct.spaceAbove + 1
+fontStruct.fontPropertiesSize := fontStruct.bitmapsTablePtr + 3
```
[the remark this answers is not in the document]
drdnar: fasmg has a struct macro? That's nice!

src/fontlibc/fontlibc.h
runer112: Yeah, I think GetLastCharacterRead fills that need fine.

src/fontlibc/fontlibc.asm
```diff
@@ -0,0 +1,1533 @@
...
+;     - true if font loaded successfully
+;     - false on failure (invalid font, or you tried to use the version byte)
+; Fetch arg0
+	ld	hl,arg0
+	add	hl,sp
+	ld	hl,(hl)
+; Verify version byte is zero like it's supposed to be
+; The literal only reason there's any validation here at all is to
+; enforce keeping the version byte reserved.
+	xor	a
```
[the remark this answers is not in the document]
drdnar: I'm used to implicit arguments, but fine, I'll change it to Zilog's redundant argument style. I guess it's more consistent with other assemblies, anyway.

src/fontlibc/fontlibc.h
drdnar: The main idea was to add a flag for ignoring the space above/below metrics. But I'm calling it `unsigned int` :p

src/fontlibc/fontlibc.h
drdnar: It's actually just a variable-size array.

. . . Speaking of, it would make more sense to put the font count immediately before the font list to serve as a list length.

runer112: If the structure isn't what I initially assumed, then it should probably be either `fontlib_font_t *font_list` (pointer to array) or ` fontlib_font_t font_list[1]` (array that can simply extend past the end of the struct).

src/fontlibc/fontlibc.h
runer112: Sure, `unsigned int` is fine too.

src/fontlibc/fontlibc.h
drdnar: I mean, the getter function is signed, so that, too, is unsigned. I'd also argue that passing a negative value makes it pretty clear you never expect the limit to be reached, whereas passing a large positive value suggests you might think it could print a huge number of characters.

src/fontlibc/fontlibc.h
drdnar: enums are ints, so then I'd either have to make the setter take an int when I only allocated a byte, or make users typecast. Which is more evil?

src/fontlibc/fontlibc.h
drdnar: Fixing

src/fontlibc/fontlibc.h
runer112: Enum constants are automatically narrowed to the target type. The signatures of the functions accepting these values would remain unchanged. But you could add doc to specify that the flags should correspond to this group of flags that now has a handy concrete name.

src/fontlibc/fontlibc.h
runer112: Passing `-1` for a value that logically represents a count implies that you're requesting some kind of special handling, but there's neither documentation nor implementation for this. And there's no reason for there to be documentation or implementations of special handling, when simply passing a value larger than can ever practically be reached (`UINT_MAX`) achieves the same purpose but without relying on special handling.

drdnar: Fair enough. But in that case, shouldn't I change `GetCharactersRemaining` it to `bool ReachedMaxCharacters(void)`?

Also, the name implies what it returns---how many characters were left before reaching `max_characters`. If the value is negative, it tells you your previous call was to the non-L variant. I suppose change Draw/Width should accept `uint24_t` but `GetCharactersRemaining` return int24_t.

src/fontlibc/fontlibc.h
drdnar: Actually, I removed it just because it looked slower than just doing a bunch of `var = GetValue();`.

I agree, however, that the struct would make for a better API.

src/fontlibc/fontlibc.h
drdnar: I'm suggesting keeping it as signed, but saying that the return value is an undefined but negative value if the last call was to the non-length-limited function variants.

src/fontlibc/fontlibc.h
drdnar: Or alternatively, have it accept a `uint16_t` for `max_characters`. The upshot is that it explicitly ensures that the return domain fully encompasses the input domain.

runer112:
> I'm not sure moving the call to `gfx_Wait` into `util.DrawGlyphRawKnownWidth` is a good move. Instead of calling it once at the start of string drawing, now it gets called for every single glyph regardless. If you do move it there, it should also be removed from `fontlib_DrawGlyph` because it's redundant.

I recognized this cost. It seemed worth it to delay the `gfx_Wait` for as long as possible, trading the extra 60 cycles per glyph drawn for waiting far less time for the first `gfx_Wait` after `gfx_SwapDraw`. But to be fair, I didn't fully consider that important last part, that this must be the first drawing call after `gfx_SwapDraw` to gain the benefit of reduced waiting.

Considering that the cost is only 60 cycles per glyph, which is probably less than 1% of the total time for drawing even a small glyph, I'd still lean towards accepting this trade. But I could be convinced otherwise.

src/fontlibc/fontlibc.h
drdnar: The reason I didn't make X and width unsigned is simply that I figured people would mostly be using `int`, signed, as it is the default. While having negative coordinates makes no sense, neither does having an X greater than 320, even though the vast majority of the domain for both `int` and `unsigned int` is outside the range [0,320).

src/fontlibc/fontlibc.h
runer112: I think I mentioned this on IRC without mentioning it anywhere on here, but I realized that the *real* type for all these values that store a number of characters should be `size_t`.

runer112:
Short of having a documented way to create fonts and write them to the proper formats, this seems good for an initial merge.

mateoconlechuga:
I think it should be rebase and merge, not squashed?

runer112:
> I think it should be rebase and merge, not squashed?

Either that or just a normal merge.

PR #178, opened 2018-12-25
update gfx_RotateSpriteC and gfx_RotateSpriteCC

grosged:
Optimised both clockwise & counter-clockwise routines

PR #189, opened 2019-03-29
new version of updated gfx_RotateSpriteC & gfx_RotateSpriteCC

grosged:
Found 2 bugs (because of LDD & LDI...) then modified both routines ;)


adriweb:
No need to open this PR, the old one https://github.com/CE-Programming/toolchain/pull/178 was automatically updated when you pushed on your fork - I'll close that one

PR #196, opened 2019-05-01
An optimisation of _PrintNormalFont

grosged:
"Unrolled" in 3 parts in order to have less comparisons inside main loop.

PR #197, opened 2019-05-01
Revert "An optimisation of _PrintNormalFont"

runer112:
Reverts CE-Programming/toolchain#196

Pretty sure this doesn't work for characters with a width other than 8. The original PR #196 really needed more review before merging.


runer112:
Also, it never should've gone straight into master. Should've gone into devel.

grosged:
Oh, I thought normal font would be 8x8...not so obvious, apparently.

grosged:
No more way to modify my mistake, now ?
(by replacing ld c,56 with ld a,64 \ sub a,ixh \ ld c,a)
I'll have to pull another request (in devel , I understood) ?

PR #198, opened 2019-05-01
Optimisation of _PrintNormalFont

grosged:
This time in devel repository ;)

### DIFF
--- a/.deploy.sh
+++ b/.deploy.sh
@@ -8,6 +8,9 @@ if [ "$TRAVIS_PULL_REQUEST" != "false" -o "$TRAVIS_BRANCH" != "master" ]; then
     exit 0
 fi
 
+# Only perform below actions on a tagged (release) commit
+git describe --exact-match --tags HEAD || exit 0
+
 rev=$(git rev-parse --short HEAD)
 
 git config user.name "Travis CI"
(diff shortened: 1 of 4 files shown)


grosged:
Sorry if not in devel : I'm afraid I did not succeeded

adriweb:
Apparently the problem from the other PR was "doesn't work for characters with a width other than 8", as mentioned by @runer112.
Does this version handle enough cases now?

adriweb:
@grosged I updated your repo, you can continue making changes to your "devel-1" branch to address reviews/comments, and the modificiations will appear in this PR :)

grosged:
@adriweb Oh thanks a lot ! Yes, I found the bug then resolved the problem of characters with width other than 8 ;)
@everyone again,  sorry for my unintentionnally bad management of github : I'm here just because I like to join assembly written projects  and nothing else, ...I have to get used to github...painfully for now =D

adriweb:
Can we get a review? If it's good, let's merge it into the LLVM branch (because llvm is the new devel...)

runer112:
The changes do all look like valid optimizations. However, I'd probably want to see a few minor adjustments (use `iy` instead of `ix` to avoid needing to save it, maybe have an early exit for fully transparent, maybe use a bit more SMC to reduce code duplication), and also want to see the spaces converted to tabs. But considering that the original PR user has been deleted, and thus their fork and the ability to make more changes to it(?), I'm rejecting this PR as-is. @grosged, I believe this was originally your PR. If you'd like to re-open it under your active account, we can revisit it.

PR #201, opened 2019-05-11
I commented out a line of code not used in program

killlmoo:


runer112:
Instead of having the weird unused line of code at all, why not just relocate the comment to the function prototypes above that include `uint8_t`?

PR #202, opened 2019-05-12
Save and restore clobbered DE register

killlmoo:
Behavior was: 
1. pqr will put "pqr\0" in the buffer
2. pqrs will put "pqrs\0" in the buffer
3. pqrst will put "pqrst" in the buffer (no NUL)
4. pqrstu will put "pqrst" in the buffer (as above, can not type the u)

### DIFF
--- a/src/ce/getstringinput.src
+++ b/src/ce/getstringinput.src
@@ -100,7 +100,9 @@ __draw:
 __full:
 	ld	a,' '
 	ld	(curUnder),a
+	push 	de
 	call	_GetKey
+	pop	de
 	cp	a,kClear
 	jr	z,__clear
 	cp	a,kEnter


adriweb:
As @mateoconlechuga mentioned, `os_GetStringInput` is kind of a hack, and eventually will be superseded by the things in #204.
In the meantime, I suppose this fix could be merged, I think at least @runer112 agreed.

(Also, I updated the PR branch target to devel as it should be)

killlmoo:
After a change has been done am I supposed to close the pull request, I am new to GitHub?

runer112:
> After a change has been done am I supposed to close the pull request, I am new to GitHub?

Nope. Just need to wait for other repository contributors to review and approve it (in this case, myself), and then for a contributor with high enough privileges to merge it (in this case, also myself).

PR #203, opened 2019-05-12
Improved comments and printTextSmall function

killlmoo:
The comment changes should make the program less confusing. The change to the printTextSmall function will enable text to be printed farther to the right, past 255 pixels.

### DIFF
--- a/examples/hello_world/src/main.c
+++ b/examples/hello_world/src/main.c
@@ -17,7 +17,7 @@
 
 /* Put function prototypes here */
 void printText(const char *text, uint8_t x, uint8_t y);
-void printTextSmall(const char *text, uint8_t xpos, uint8_t ypos);
+void printTextSmall(const char *text, uint16_t xpos, uint8_t ypos);
 
 /* Put all your code here */
 void main(void) {
--- a/examples/hello_world/src/main.c
+++ b/examples/hello_world/src/main.c
@@ -10,16 +10,17 @@
 #include <stdlib.h>
 #include <string.h>
 
+/* uint8_t is an unsigned integer that can range from 0-255. */
+/* It performs faster than just an int, so try to use it (or int8_t) when possible */
+/*In this program uint8_t is used as the main data type*/
+
+
 /* Put function prototypes here */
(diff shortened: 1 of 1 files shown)


adriweb:
Looks good except that line 20 also has to be updated, as we can see by the build failures.

killlmoo:
> Looks good except that line 20 also has to be updated, as we can see by the build failures.

Yep working on that, also trying to learn how to use Github (I am new).

adriweb:
No need to close and reopen PRs each time, when you commit on your branch (where the PR is from), new commits will appear here as well.

killlmoo:
> No need to close and reopen PRs each time, when you commit on your branch (where the PR is from), new commits will appear here as well.

Good to know, I am lost on getting back to my branch to edit it though.

adriweb:
There's a link at the top of the page :)

https://github.com/killlmoo/toolchain/tree/patch-1 in this case

killlmoo:
> There's a link at the top of the page :) https://github.com/killlmoo/toolchain/tree/patch-1 in this case

Figured it out, thanks :).

mateoconlechuga:
I'm rejecting this pr currently because of the use of uint16_t. Please fix that. Also `printTextSmall` should just be removed from the program, it is a worthless function.

adriweb:
BTW, the target branch was changed to "devel" because we use that for all "development" purposes (we try to use the "master" branch to reflect a stable state of things).

And about `uint16_t`, it turns out it's not quite recommended if I understood Mateo correctly, but yeah, the whole `printTextSmall` function can go away anyway :P

killlmoo:
> I'm rejecting this pr currently because of the use of uint16_t. Please fix that. Also `printTextSmall` should just be removed from the program, it is a worthless function.

The `printTextSmall` function needed the uint16_t because the pixel width of the screen is bigger than 255, it is 320. Anyways, the useless function was removed in the end.

mateoconlechuga:
Sorry, I guess I should be more clear. `uint16_t` isn't recommended for use on the ez80 because it isn't a word size, thus increasing the amount of generated code.

PR #225, opened 2019-09-25
Elaboration on the scaling-related functions

Zexxerd:
More info on the scaling-related functions was added, and some typos.

I was only able to gather this information through empirical observations, by the way.


mateoconlechuga:
This doesn't make any sense... why is 64 even mentioned? This just confuses people more.

mateoconlechuga:
Added with 055460b510c846fa654518ca0806461aed46baa7

PR #230, opened 2020-01-11
merge into devel the commits from llvm branch without llvm-specific stuff

adriweb:


runer112:
Some overall comments:

- Why did we change away from OTB? If I'm the only one that this bothers, I'll suck it up, but if this was just a Mateo being Mateo thing and others don't like it either, please speak up.
- Why are generated files from convpng (convimg?) committed? They're build artifacts.

examples/graphx/floodfill/src/main.c
```diff
@@ -26,7 +26,7 @@ int main(void)
     gfx_Begin();
 
     /* Use the default palettes, this is already set by gfx_Begin */
-    gfx_SetDefaultPalette();
+    gfx_SetDefaultPalette(gfx_8bpp);
```
runer112: This line should just be removed, it does nothing in this context.

adriweb: done

adriweb:
So, the toolchain builds , and so far from running the built examples, everything seems to be OK except examples/combined/appvar_tilemap which just resets the calc (the autotest failed as expected)

mateoconlechuga:
what is OTB

adriweb:
> what is OTB

A brace style: https://softwareengineering.stackexchange.com/a/99557

adriweb:
Cleaned up the last commits, Travis and Appveyor are happy now too.
I believe this is mergeable into devel.

PR #233, opened 2020-01-12
Fix defines

runer112:
Add parentheses and volatile where necessary.


runer112:
Someone please confirm whether or not the things made volatile make sense. In particular, `lcd_Ram` and `lcd_Palette` may not.

mateoconlechuga:
Looks good

PR #245, opened 2020-03-10
Implement argc/argv, add an example/test program

runer112:


runer112:
I also optimized the BSS clearing code while I was in cstartup, including hand-optimized special cases for every length that can do better than the general case (which was also optimized a bit).

runer112:
So... anyone want to review this..?

PR #248, opened 2020-03-16
Update graphx.asm

grosged:
Optimisation of gfx_SetTextXY

### DIFF
--- a/src/graphx/graphx.asm
+++ b/src/graphx/graphx.asm
@@ -2967,13 +2967,15 @@ gfx_SetTextXY:
 ;  arg1 : Text Y Pos
 ; Returns:
 ;  None
-	pop	de			; de=return address, sp=&xpos
-	pop	hl			; hl=xpos, sp=&ypos
-	ld	(_TextXPos),hl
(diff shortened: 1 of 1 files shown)


runer112:
Clipped text is supported, so X and Y need to be full 24-bit values.

adriweb:
(Reopening to address feedback)

runer112:
I don't believe that this optimization can really be reworked to still apply, as the existing code already looks optimal for full 24-bit values.

grosged:
oh sorry : I did not realised clipped text would be supported.
Full 24-bit for _TextXpos, okay !.. But, concerning _TextYPos, these lines make me think "8 bit" :
ld	hl,0
_TextYPos := $-3
	ld	h,LcdWidth/2
	mlt	hl

runer112:
Yes, that's the unclipped version of the text printing routine. If you look [here](https://github.com/grosged/toolchain/blob/496c8e0f713246b813b473a194d3c24799974c32/src/graphx/graphx.asm#L3302) at the clipped version, though, it reads the full 24-bit X and Y values to pass them on to `gfx_TransparentSprite`.

PR #249, opened 2020-03-16
Update graphx.asm

grosged:
Optimisation of gfx_PrintInt


runer112:
This portion of code was intentionally optimized for size rather than speed, because the speed difference here is so small compared to the total time the function will take.

PR #250, opened 2020-03-17
Update graphx.asm

grosged:
Here's my new version of _PrintNormalFont (faster & with a few self-modifying-codes)
Could you please check & review ?.. Thx !

### DIFF
--- a/src/graphx/graphx.asm
+++ b/src/graphx/graphx.asm
@@ -3177,34 +3177,58 @@ smcByte _TextHeight
 	jr	_PrintLargeFont		; SMC the jump
 _TextScaleJump := $ - 1
 _PrintNormalFont:
-.loop:
-	ld	c,(hl)			; c = 8 pixels
-	add	iy,de			; get draw location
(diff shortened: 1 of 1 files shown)


runer112:
From looking at the code, it all looks good. @grosged have you tested any of the text-printing examples with these optimizations to verify that they still work? I'll probably do so myself, just wondering if you have as well.

runer112:
Could you convert spaces to (8-wide) tabs, though? Or maybe I can if I have that permission, not sure.

grosged:
@runer112,
I verified nothing : I  just wrote & re-adapted/shortened these optimizations in my 'mind' only.
I'll try to convert to good 8 wide tabs ;)

grosged:
Unsuccesfully modified : I'm using github "online" within my browser (not the best manner, I know) I just can't change into 8 space wide tabs, sorry.
I'm  a linux user, so no github desktop for me :(

grosged:
Thanks, jacobly0 ;)

PR #251, opened 2020-03-25
Optimisation of kb_AnyKey (shorter/faster)

grosged:


runer112:
Untested, I'm guessing?

runer112:
@mateoconlechuga @jacobly0 Why don't we use mode 1 for this anyway?

mateoconlechuga:
Of course it hasn't been tested. I say we don't accept pull requests anymore until he starts testing them.

mateoconlechuga:
To add on to that, it seems sketchy as heck to rely on port mirroring like this.

adriweb:
In addition, we unfortunately don't have any autotests for `kb_AnyKey` yet.
_(That said, there are only 2 tests left that don't pass, I'll try to see what's going on)_

adriweb:
To re-open later if needed (especially if we can make sure the "scary port things" are actually safe)

adriweb:
jk. i've been told this is already used elsewhere.
Merged as 4eb48e9a9a035c9808205b00965cdeab2798a9ef

PR #258, opened 2020-05-03
Add ti::real C++ class wrapping TI's real number format and OS funcs

adriweb:
**TODO**:

 * [x] Better math-oriented example?
 * [x] Autotest on that example
 * [x] wrap the remaining functions


adriweb:
We could (should) probably come up with a fancy math example though.

erentar:
Please put the monte carlo pi as math example

adriweb:
Sure, that's easy enough

adriweb:
OK, review plz, I added more functions and tests.

src/ce/tireal.hpp
```diff
@@ -0,0 +1,585 @@
+#pragma once
```
mateoconlechuga, line 1: no

src/ce/tireal.hpp
```diff
@@ -0,0 +1,585 @@
...
+
+#include <tice.h>
+
+#include "limits.hpp"
```
mateoconlechuga, line 5: should be the very first include

src/std/limits.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
```
mateoconlechuga, line 1: no

PR #259, opened 2020-05-30
Fix typo in 'Merthsoft'

merthsoft:


runer112:
What, you don't want to be known as "Methsoft"?

runer112:
This change is obviously fine, but I'm a bit confused by our branch strategy at the moment so I'll hold off until someone more knowledgeable sees this and can verify what the target branch should be.

merthsoft:
Fair! Happy to target a different branch if needed :)

adriweb:
Yeah basically the only real development branch now is llvm, that's what the target should be. I'll change it and merge the PR soon.

merthsoft:
ok I made a mess changing the source. I can make a different branch off of llvm

PR #267, opened 2020-07-13
Improve fileioc garbage collect handling

beckadamtheinventor:
When a gc will be triggered as a result of calling _Arc_Unarc, set OS graphics mode, restore afterwards.
Added ti_SetPostGCHandler((void)(*routine)(void)) so user programs can easily setup graphics again.
Had to replace some of the files in ``src/include/`` to get fileioc to build. Ignore those if you can lol
[archive_overloader.zip](https://github.com/CE-Programming/toolchain/files/4914240/archive_overloader.zip)


adriweb:
Woops, please reopen targeting the `llvm` branch, yep :)

PR #269, opened 2020-07-13
Improve fileioc garbage collect handling

beckadamtheinventor:
When a gc will be triggered as a result of calling _Arc_Unarc, set OS graphics mode, restore afterwards, then run a callback.
Default callback sets up xlibc palette and zeroes the screen. Use `ti_SetPostGCHandler((void)(*routine)(void))` to change this callback.
(Third PR because I keep screwing up the PR location)
[archive_overloader.zip](https://github.com/CE-Programming/toolchain/files/4914366/archive_overloader.zip)


beckadamtheinventor:
Hopefully fileioc.asm doesn't disappear again...

adriweb:
See #270

Issue #222, opened 2019-08-09
Garbage Collect


beckadamtheinventor:
Garbage collect should be a standard Fileioc function.
I would also suggest a function that would tell you when a garbage collect would be necessary.
I assume this is just about calculating the largest free memory space in archive, let me know if I'm wrong.

mateoconlechuga:
Garbage collect should not be a fileioc function.

But I added `ti_ArchiveHasRoom` so you can check if it triggers.

runer112:
Manual garbage collection has merit and should be relatively easy to implement.

PR #270, opened 2020-07-13
fileioc: improve garbage collect handling

beckadamtheinventor:
When a gc will be triggered as a result of calling _Arc_Unarc, set OS graphics mode, restore afterwards.
Added ti_SetPostGCHandler((void)(*routine)(void)) so user programs can easily setup graphics again.
Had to replace some of the files in src/include/ to get fileioc to build. Ignore those if you can lol
[archive_overloader.zip](https://github.com/CE-Programming/toolchain/files/4914759/archive_overloader.zip)

### DIFF
--- a/src/fileioc/fileioc.asm
+++ b/src/fileioc/fileioc.asm
@@ -1228,7 +1228,7 @@ ti_Rename:
 	jr	c, .return_2
 	call	_ChkInRam
 	jr	nz, .in_archive
-	ld	hl, $f8			; $f8 = ret
+	ld	hl, util_no_op			; no-op routine instead of assuming $F8 points to a ret instruction lol
 	ld	(.smc_archive), hl
 	call	_PushOP1
 	call	util_Arc_Unarc
@@ -1419,7 +1419,7 @@ ti_SetPostGCHandler:
 	ld (util_post_gc_handler),hl
 	ex hl,de
 	jp (hl)
-util_post_gc_default_handler:=$F8
+util_post_gc_default_handler:=util_no_op
 
 ;-------------------------------------------------------------------------------
 ti_SetPreGCHandler:
@@ -1442,6 +1442,7 @@ ti_SetPreGCHandler:
 	jp (hl)
 util_pre_gc_default_handler:
 	xor a,a
+util_no_op:
 	ret
 
 
--- a/src/fileioc/fileioc.asm
+++ b/src/fileioc/fileioc.asm
@@ -1440,7 +1440,9 @@ ti_SetPreGCHandler:
 	ld (util_pre_gc_handler),hl
 	ex hl,de
 	jp (hl)
-util_pre_gc_default_handler:=$F8
+util_pre_gc_default_handler:
+	xor a,a
+	ret
 
 
 ;-------------------------------------------------------------------------------
@@ -1636,6 +1638,8 @@ util_Arc_Unarc: ;properly handle garbage collects :P
 	jr nz,.arc_unarc ;gc will not be triggered
 	call util_pre_gc_default_handler
 util_pre_gc_handler:=$-3
+	or a,a
+	ret nz ;exit if the handler returns a non-zero value
 	call _Arc_Unarc
 	jp util_post_gc_default_handler
 util_post_gc_handler:=$-3
--- a/src/fileioc/fileioc.asm
+++ b/src/fileioc/fileioc.asm
@@ -60,6 +60,13 @@ library 'FILEIOC', 5
 	export ti_ArchiveHasRoom
 
 ;-------------------------------------------------------------------------------
+; v6 functions
+;-------------------------------------------------------------------------------
+	export ti_SetPostGCHandler
+
+
+
+;-------------------------------------------------------------------------------
 vat_ptr0 := $d0244e
 vat_ptr1 := $d0257b
 vat_ptr2 := $d0257e
@@ -461,7 +468,7 @@ ti_SetArchiveStatus:
 .set_archived:
 	push	bc
 	pop	af
-	call	z, _Arc_Unarc
+	call	z, util_Arc_Unarc
 	jr	.relocate_var
 .set_not_archived:
 	push	bc
@@ -1206,7 +1213,7 @@ ti_Rename:
 	inc	de
 	call	_Mov8b
 	call	_PushOP1		; save old name
-	ld	hl, _Arc_Unarc
+	ld	hl, util_Arc_Unarc
 	ld	(.smc_archive), hl
 	pop	hl			; new name
 	ld	de, OP1 + 1
@@ -1224,7 +1231,7 @@ ti_Rename:
 	ld	hl, $f8			; $f8 = ret
 	ld	(.smc_archive), hl
 	call	_PushOP1
-	call	_Arc_Unarc
+	call	util_Arc_Unarc
 	call	_PopOP1
 	jr	.locate_program
 .in_archive:
@@ -1257,7 +1264,7 @@ ti_Rename:
 	ldir
 .is_zero:
 	call	_PopOP1
-	call	_Arc_Unarc
+	call	util_Arc_Unarc
 .smc_archive := $-3
 	call	_PopOP1
 	call	_ChkFindSym
@@ -1392,6 +1399,26 @@ ti_ArchiveHasRoom:
 	ret	nz
 	dec	a
 	ret
+
+;-------------------------------------------------------------------------------
+ti_SetPostGCHandler:
+;Set handler for setting up the screen after a garbage collect
+; args:
+;   sp + 3 : pointer to handler. Set to 0 to use default handler (xlibc palette)
+; return:
+;   None
+	pop de
+	ex (sp),hl
+	push de
+	add hl,de
+	or a,a
+	sbc hl,de
+	jr nz,.notdefault
+	ld hl,util_gfx_restore_default_handler
+.notdefault:
+	ld (util_gfx_restore_handler),hl
+	ex hl,de
+	jp (hl)
 
 ;-------------------------------------------------------------------------------
 ; internal library routines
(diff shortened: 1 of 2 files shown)


adriweb:
Kinda related to #222

src/fileioc/fileioc.asm
```diff
@@ -1573,6 +1600,67 @@ util_set_offset:
...
+	call _FindFreeArcSpot ;check if we will trigger a gc
+	jr nz,.arc_unarc ;gc will not be triggered
+	ld a,(mpLcdCtrl)
+	cp a,lcdBpp16
+	jr z,.arc_unarc ;already in OS 16bpp graphics mode
+	push af ;save lcd mode
+	call _ClrLCDFull
+	call _DrawStatusBar
+	ld a,lcdBpp16
+	ld (mpLcdCtrl),a
```
runer112, line 1621: Should become a call to the default, no-op pre-GC handler.

src/fileioc/fileioc.asm
```diff
@@ -1573,6 +1600,67 @@ util_set_offset:
...
+	call _Arc_Unarc
+	pop af ;restore lcd mode
+	ld (mpLcdCtrl),a
+	jp util_gfx_restore_default_handler
```
runer112, line 1625: Should become a jump to the default, no-op post-GC handler.

src/fileioc/fileioc.asm
```diff
@@ -1573,6 +1600,67 @@ util_set_offset:
...
+	ld	a,31
+	and	a,b
+	or	a,e
+	ld	(hl),a
+	inc	hl
+	ld	(hl),d
+	inc	hl
+	inc	b
+	jr	nz,.loop
+	ret
```
runer112, line 1662: This GraphX-specific default handler should go away.

src/fileioc/fileioc.h
```diff
@@ -382,6 +382,14 @@ uint8_t ti_RclVar(const uint8_t var_type, const char *var_name, void **data_stru
  */
 bool ti_ArchiveHasRoom(uint24_t num_bytes);
 
+
+/**
+ * Set routine to run after a garbage collect.
+ * @param routine Routine to run following a garbage collect. If this is 0, the default handler will be used.
```
runer112, line 388: Once the default handler is a no-op, it's probably clearer to just say that 0 will clear any handler set.

adriweb: Might as well say `NULL` instead of `0`, too.

src/fileioc/fileioc.asm
beckadamtheinventor: Default handler now points to a ret instruction, added pre-GC handler.

src/fileioc/fileioc.asm
beckadamtheinventor: Fixed.

beckadamtheinventor:
Changed return type required for pre-gc handler. If it returns a non-zero value, the archival which causes the garbage collect will be cancelled.

src/fileioc/fileioc.asm
```diff
@@ -1393,6 +1400,51 @@ ti_ArchiveHasRoom:
...
+	add hl,de
+	or a,a
+	sbc hl,de
+	jr nz,.notdefault
+	ld hl,util_post_gc_default_handler
+.notdefault:
+	ld (util_post_gc_handler),hl
+	ex hl,de
+	jp (hl)
+util_post_gc_default_handler:=$F8
```
runer112, line 1422: What's going on here? Shouldn't this label simply point to a `ret`? (And you can reuse a preexisting one elsewhere, like the one in `util_pre_gc_default_handler`.)

src/fileioc/fileioc.h
```diff
@@ -382,6 +382,38 @@ uint8_t ti_RclVar(const uint8_t var_type, const char *var_name, void **data_stru
...
+void ti_SetPostGCHandler(void (*routine)(void));
+
+/**
+ * Set routine to run before a garbage collect.
+ * @param routine Routine to run preceeding a garbage collect. NULL sets it to do nothing. If this routine returns true, the variable will not be archived.
+ * @note Useful for cleanup. If your program uses graphx, run gfx_End inside the passed function to set OS graphics mode to avoid corrupted graphics during the garbage collect.
+ * @code
+ *    bool pre_gc_handler(void){
+ *          gfx_End();
+ *          return 0;
```
runer112, line 400: I was about to leave a comment that this documentation would need to be updated like this, but it looks like you were already on top of it. The one change I'd request, though, is that `0` be changed to `false`.

src/fileioc/fileioc.h
```diff
@@ -382,6 +382,38 @@ uint8_t ti_RclVar(const uint8_t var_type, const char *var_name, void **data_stru
...
+/**
+ * Set routine to run after a garbage collect.
+ * @param routine Routine to run following a garbage collect. NULL sets it to do nothing.
+ * @note If your program uses graphx, pass gfx_Begin to setup graphics after the garbage collect finishes.
+ * */
+void ti_SetPostGCHandler(void (*routine)(void));
+
+/**
+ * Set routine to run before a garbage collect.
+ * @param routine Routine to run preceeding a garbage collect. NULL sets it to do nothing. If this routine returns true, the variable will not be archived.
```
runer112, line 395: It should probably also be mentioned that, if the routine returns `true`, the post-GC handler will not be run.

runer112:
By the way, although I'm bombarding you with change requests, I'd like to state that I appreciate your work on this. After my last few comments are resolved, it looks like everything should be in good shape from my perspective.

src/fileioc/fileioc.asm
beckadamtheinventor: $F8 points to a ret instruction.

runer112: On your OS/boot code version, sure, but there's no guarantee it's that way for others. There's really no reason not to point to an internal `ret` instead.

beckadamtheinventor: Alright, will update this routine as well as the others that use $F8 as a no-op :P

src/fileioc/fileioc.h
beckadamtheinventor: Actually, should I set the post-gc handler to run regardless of a gc triggering?

runer112: I was thinking about that, but if the pre-GC handler returns `true`, it makes sense to just do that before making any changes in preparation for a GC that's not going to happen that would need to be reverted by the post-GC handler.

beckadamtheinventor: Fair enough.

src/fileioc/fileioc.h
beckadamtheinventor: Edited

src/fileioc/fileioc.asm
beckadamtheinventor: Resolved with latest commit

beckadamtheinventor:
All comments should be resolved. I think there's another comment, but I don't see it lol.

mateoconlechuga:
Looks nice enough. Can you just fix the spacing please?

beckadamtheinventor:
Done. :P

src/fileioc/fileioc.asm
```diff
@@ -1573,6 +1626,29 @@ util_set_offset:
...
+	jr	nz,.arc_unarc ;if the file is already in archive, we won't trigger a gc
+	call	_LoadDEInd_s
+	ld	hl,12
+	add	hl,de
+	call	_FindFreeArcSpot ;check if we will trigger a gc
+	jr	nz,.arc_unarc ;gc will not be triggered
+	call	util_pre_gc_default_handler
+util_pre_gc_handler:=$-3
+	or	a,a
+	ret	nz ;exit if the handler returns a non-zero value
```
runer112, line 1643: One last comment, I swear.

Do people think it would make more sense for this condition to be reversed? Previously, the return type was an integer, in which case it made sense for non-zero to signal error/abort. But now that the return type is a `bool`, I feel like it makes more sense for `false` (zero) to signal error/abort.

mateoconlechuga: I agree that should probably be a ret z. The docs would have to be updated as well.

beckadamtheinventor: Fixed and updated.

mateoconlechuga:
I'm good with a merge

src/fileioc/fileioc.asm
```diff
@@ -1573,6 +1626,29 @@ util_set_offset:
 	ld	(hl), bc
 	ret
 
+util_Arc_Unarc: ;properly handle garbage collects :P
+	call	_ChkFindSym
+	push	hl
+	call	_ChkInRAM
+	pop	hl
+	jr	nz,.arc_unarc ;if the file is already in archive, we won't trigger a gc
+	call	_LoadDEInd_s
```
jacobly0, line 1635: There is already a routine that does this.

src/fileioc/fileioc.asm
```diff
@@ -1573,6 +1626,29 @@ util_set_offset:
...
+	ld	hl,12
+	add	hl,de
+	call	_FindFreeArcSpot ;check if we will trigger a gc
```
jacobly0, line 1638: There is already a routine that does this.

src/fileioc/fileioc.asm
```diff
@@ -1573,6 +1626,29 @@ util_set_offset:
...
+	jr	nz,.arc_unarc ;gc will not be triggered
+	call	util_pre_gc_default_handler
+util_pre_gc_handler:=$-3
+	or	a,a
+	ret	z ;exit if the handler returns false
```
jacobly0, line 1643: This is a fail state (the caller either failed to do what it is trying to do or will soon do bad things) which needs to be explicitly handled by the caller.  Especially if there were a global state for whether to allow garbage collection, ti_SetArchiveStatus needs to return whether it actually succeeded.

src/fileioc/fileioc.asm
```diff
@@ -1393,6 +1400,52 @@ ti_ArchiveHasRoom:
 	dec	a
 	ret
 
+;-------------------------------------------------------------------------------
+ti_SetPostGCHandler:
+;Set handler for setting up the screen after a garbage collect
+; args:
+;   sp + 3 : pointer to handler. Set to 0 to use default handler (xlibc palette)
```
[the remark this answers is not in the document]
beckadamtheinventor: Sounds good, I'll update that.

src/fileioc/fileioc.asm
beckadamtheinventor: optimized.

beckadamtheinventor: hmm

src/fileioc/fileioc.asm
beckadamtheinventor: optimized, now uses it

src/fileioc/fileioc.asm
beckadamtheinventor: removed gc canceling

src/fileioc/fileioc.asm
```diff
@@ -1393,6 +1400,52 @@ ti_ArchiveHasRoom:
...
+	ld	(util_post_gc_handler),hl
+	ex	hl,de
+	jp	(hl)
+util_post_gc_default_handler:=util_no_op
+
+;-------------------------------------------------------------------------------
+ti_SetPreGCHandler:
+;Set handler for setting up the screen after a garbage collect
+; args:
+;   sp + 3 : pointer to handler. Set to 0 to use default handler (xlibc palette)
```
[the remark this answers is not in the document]
beckadamtheinventor: fixed

src/fileioc/fileioc.asm
beckadamtheinventor: Using a subroutine wouldn't help size nor speed.

jacobly0: How is avoiding a _ChkFindSym possibly "not helping speed".

beckadamtheinventor: oops, didn't realise it wasn't needed there

beckadamtheinventor: updated

adriweb:
Review?

src/fileioc/fileioc.asm
```diff
@@ -1393,6 +1400,36 @@ ti_ArchiveHasRoom:
...
 
+;-------------------------------------------------------------------------------
+ti_SetGCBehavior:
+;Set routines to run before and after a garbage collect would be triggered.
+; args:
+;   sp + 3 : pointer to routine to be run before. Set to 0 to use default handler.
+;	sp + 6 : pointer to routine to be run after. Set to 0 to use default handler.
+; return:
+;   None
+	pop	de
```
mateoconlechuga, line 1411: can someone else just double check that this pop order is correct? it seems that it sets up the pointers backwards but maybe I'm just tired.

beckadamtheinventor: yeah that apparently changed during my last pull lol. Fixed it

beckadamtheinventor: actually, no. Just realised that was an intentional optimization

mateoconlechuga: I'm aware. I don't think it is ever correct; which is why I am asking.

beckadamtheinventor: see here  https://github.com/beckadamtheinventor/toolchain/blob/a6118329b37020a9790a18bffaf8058e630b3f42/src/fileioc/fileioc.asm#L1424

mateoconlechuga: Dude. I am asking you to look at if you are popping the arguments in the correct order. I know how to write code.

adriweb:
well the CI is still in progress, but the toolchain builds...
Anyway, eventually it would be nice to add an autotested example too :)

PR #272, opened 2020-07-28
Added TextIOC library and demos

Captain-Calc:
Here is the library! Please let me know if you have any recommendations.


examples/textioc/advanced-input-features/src/fonts/fonts.c
```diff
@@ -0,0 +1,11 @@
+#include "fonts.h"
+
+/* This contains the raw data for the font. */
+static const uint8_t test_font_data[] = {
+	#include "testfont.inc"
+};
+
+/* However, C89 does not allow us to typecast a byte array into a
```
adriweb, line 8: LLVM is being targeted, you don't need to care about C89 anymore :)

adriweb:
The Travis build is failing because the lib isn't actually being included/built as part of the toolchain.
You need to add it there to make everything know about it:

- https://github.com/CE-Programming/toolchain/blob/llvm/makefile#L6
- https://github.com/CE-Programming/toolchain/blob/llvm/tools/doxygen/config#L169
- https://github.com/CE-Programming/toolchain/blob/llvm/tools/doxygen/config#L800
- https://github.com/CE-Programming/toolchain/blob/llvm/CMakeLists.txt#L10

adriweb:
If you know how to, please ask some autotests :)
If you don't, feel free to ask us how to make those.

Captain-Calc:
Just pushed a new commit to clean up some typos before I saw your comment. I will make the necessary changes to the makefiles and configs. I do not know how to autotest. Could you explain, please? :)

Captain-Calc:
I pushed the 'Test version 2.0' commit before I saw the suggestion about rebasing the toolchain on the Cemetech SAX. Unfortunately, the message had disappeared before I could read it more carefully. Which rebase command do you suggest?

jacobly0:
Click on the red x right above to see the current build failures that block this PR being merged.

Captain-Calc:
I'll rebase the PR, build the examples, and force push the changes

mateoconlechuga:
No don't do anything yet

Captain-Calc:
Ok

src/textioc/textioc.asm
```diff
@@ -0,0 +1,3970 @@
...
+	
+
+
+
+;-------------------------------------------------------------
+; Internal Library Functions
+;-------------------------------------------------------------
+
+
+util.CallHL:
```
mateoconlechuga, line 3532: this should probably check that hl != 0 rather than just blindly jumping if the user is dumb

Captain-Calc: Agreed. I will fix that and update the documentation tomorrow.

Captain-Calc:
The last commit fixed the example build problems. I put all of the examples in the textioc/ directory and prefixed them with the library that they used. This can be a temporary solution for build testing. I do not know why the secrets part of the build failed.

adriweb:
You can fetch llvm and rebase now. any modification done to files that aren't yours, you can take as-is, in case of conflicts when you rebase (which should happen since there were some force-pushes done at some point).

Captain-Calc:
While I was struggling with the git commands, I accidentally fetched all of the toolchain branches, so I believe that my local repository is now corrupted. Should I delete it and re-clone the LLVM branch?

adriweb:
`git fetch --all` (if that's what you mean) doesn't "corrupt" your repo though :P

If you don't have any non-committed change, try this: `git clean -xfd` then `git reset --hard 545fe23`, assuming you're on your llvm branch.

Captain-Calc:
git fetch upstream/llvm and git rebase upstream/llvm both return fatals even though https://www.github.com/CE-Programming/toolchain.git' was both the upstream fetch and push target.

src/textioc/textioc.asm
```diff
@@ -0,0 +1,3975 @@
...
+	include '../include/include_library.inc'
+
+;-------------------------------------------------------------
+	library 'TEXTIOC',2
+;-------------------------------------------------------------
+
+;-------------------------------------------------------------
+; Dependencies
+;-------------------------------------------------------------
+	include_library '../graphx/graphx.asm'
```
mateoconlechuga, line 14: Why is there still a dependency on graphx?

Captain-Calc: I use gfx_FillRectangle_NoClip to draw the cursor in textio_Input

mateoconlechuga: What if I want to use this library on the homescreen to get text input?

Captain-Calc: I remember looking at the homescreen cursor functions to see if TextIOC could support it, and I believe it lacked some of the necessary functions. I might be misremembering, however.

mateoconlechuga: You shouldn't have to do much for homescreen support. Just provide wrappers to the functions that you need; and abstract them. Don't bother writing an example or trying to implement a homescreen example, that's for the user to do. Keep in mind since you are using GetCSC, input is already horridly slow, so a few clocks makes zero difference.

Captain-Calc: Although I would like to make *TextIOC* as extensible as possible, implementing homescreen routines would require resolving a great deal of compatibility issues, including translating pixel coordinates to row-column coordinates. The amount of time I have to do calculator programming is also going to diminish substantially for an indefinite period in the next few days, so I would like to complete this version although I understand that if I do implement them in a version 3, there might be some chaos regarding new programs using outdated *TextIOC* versions. If you would like, you are more than welcome to add homescreen support to *TextIOC* either in this version or a new one.

mateoconlechuga: All right fair enough :)

adriweb: Keep it in your todolist for the next version, also writing somewhere (readme?) as a caveat that homescreen stuff isn't supported yet :P

Anyway, is the current state ready to merge?

Captain-Calc: I will do that. :) I will look over the examples and documentation one last time to check for any typos or any documentation that did not get updated during the development process.

Captain-Calc:
All done and ready!

mateoconlechuga:
All right, so I went through a little bit and here are my initial thoughts:

1) These are function pointers, not random void pointers. They should be typed appropriately: https://github.com/CE-Programming/toolchain/blob/2850d9fab111df76b53bebc013e46bf4d049a4bd/src/textioc/textioc.h#L96-L103

2) This should be a defined macro: https://github.com/CE-Programming/toolchain/blob/2850d9fab111df76b53bebc013e46bf4d049a4bd/src/textioc/textioc.h#L159

3) I fail to see the reasoning behind forcing dynamic allocation of all structures. It's very wasteful on limited resources, and the user should be managing this, not the library. Instead, functions such as `textio_AllocIDS` and `textio_CreateIDS` should use a user-supplied structure and initialize it, and not perform any allocation of memory.

4) The `uint24_t *` type is used to pass around the many different structure types in this library. That makes no sense. This should be a typedef'd or opaque pointer. typedef is preferable because of 5. However, an internal opaque structure could also be used.

5) There are many functions in the library that just get data from the structure, e.g. `textio_GetCurrCharPtr`, `textio_GetIDSX`, `textio_GetIDSY`. I am under the impression that a C structure should be presented to the user instead, and the ability to modify parameters would be encased in this structure. Macros could be used if warranted to set elements in the structure. Using functions for this sort of thing slows things down and adds necessary overhead and size.

6) I don't see the purpose of timed inputs being a part of the library. It restricts the user from using a timer, 

7) There should not be any theta-specific handling functions. This includes `textio_GetThetaCodepoint`, `textio_SetDrawThetaCharFunction`, `textio_ConvertProgramAppvarName_TIOS` and others. It doesn't make sense because it is trying to workaround an issue that is controlled by a different library, or by the user themseleves. The user has control over the characters and settings of the other libraries, and this library should have no knowledge of this.

9) The windowing function have the same issues described 5.

10) `textio_PrintText`, `textio_GetStringWidthL`, `textio_GetLineWidth`, and any other function that takes a user string/character should be `const`.

8) Limiting to `os_GetCSC` input means input cannot be received in a time-critical loop. This is fine for now; but should probably be given more thought in future revisions of the library.
11) There's currently no way to use this outside of 8bpp mode. If a user wants to do something special, they can't use this library.
https://github.com/CE-Programming/toolchain/pull/272/files#diff-a8a33a82081e8798cd578338b862d459R212
https://github.com/CE-Programming/toolchain/pull/272/files#diff-a8a33a82081e8798cd578338b862d459R270
https://github.com/CE-Programming/toolchain/pull/272/files#diff-a8a33a82081e8798cd578338b862d459R323
https://github.com/CE-Programming/toolchain/pull/272/files#diff-a8a33a82081e8798cd578338b862d459R738
https://github.com/CE-Programming/toolchain/pull/272/files#diff-a8a33a82081e8798cd578338b862d459R747
https://github.com/CE-Programming/toolchain/pull/272/files#diff-a8a33a82081e8798cd578338b862d459R729
https://github.com/CE-Programming/toolchain/pull/272/files#diff-a8a33a82081e8798cd578338b862d459R694

Captain-Calc:
I have made some progress during the past week on TextIOC. My goal is to make the library responsible only for memory manipulation (e.g. inserting characters in a text buffer, removing them, etc.). How the contents of the buffer are displayed is up to the programmer. The automated input function has been completely removed, so the programmer can create his own unique input routine using the basic memory routines that the library now provides. I am currently experimenting with multi-line input.

Captain-Calc:
What do you gentlemen think of the [current library](https://www.github.com/Captain-Calc/textioc)? Are there any features you would like to see in it, bugs that should be fixed, etc.?

mateoconlechuga:
I would like to see multiple windowing support. It seems that right now everything is global in the library.

Captain-Calc:
Do you mean the PrintFormat and Newline/Tab characters?

Captain-Calc:
Is there a resource for how structures work in eZ80 assembly? I did some experimenting with DrDnar's code in FontLibC but didn't figure out how they worked.

The version that allowed custom keymaps essentially allowed the user to make variations of the os_GetCSC string example. With this version, all the user would have to do to swap keymaps is change the string that they pass to textio_KeyToOffset.

I am also a little proud of that function because it was in its design that I figured out how key input works and because it was such a small function. :) But if it's decided that it is not useful, it can go.

Captain-Calc:
I did notice while writing the examples that several functions like deleting/inserting characters, moving the cursor, etc. seemed fairly standard between GraphX, FontLibC, and TI-OS. I though it might be worth it to create some default library functions that would take care of these so if the user just wanted a basic input routine, they could whip one up in under 1000 bytes.

Captain-Calc:
I created separate default output data structures for FontLibC, GraphX, and the homescreen to give the user more default settings that they could use. I also created a prototype "GetChar" routine. To make the implementation a little easier, I bundled the buffer pointer, buffer size, char pointer, etc. into a structure. The routine is essentially the C block commented out above in the GraphX [auto-scrolling-input](https://github.com/Captain-Calc/textioc/blob/ca986afc19b07255f9a575266387da9de21af439/examples/graphx/auto-scrolling-input/src/main.c#L94) example.

mateoconlechuga:
All right, so I am sorry if this comment comes off as rude, but there are numerous things that don't make sense to me:

1) What is the point of this library anymore? Every function in it is merely an single assignment or combination of memcpy/memmove/memet. Why does there need to be a library for working with buffers, when the C standard library already provides this?

2) Why did you completely change the functionality? There was nothing wrong with what it was before, and it was useful. Now someone has to implement their own input and printing routines? For example, this was previous handled by the library, but now it is all on the user to do it? https://github.com/Captain-Calc/textioc/blob/ca986afc19b07255f9a575266387da9de21af439/examples/fontlib/wrap-text/src/main.c#L25-L63

3) The documentation doesn't really tell me anything. For example, in [this](https://github.com/Captain-Calc/textioc/blob/ca986afc19b07255f9a575266387da9de21af439/textioc.h#L105-L114), what is location? What is buffer? What does "Insert" mean vs. "Write"? How is the user supposed to know what these functions do and how to use them?

4) I feel like all these issues could be solved by establishing an API that is agreed upon and makes sense. I thought we were pretty close with the previous library, but this new one I am at a loss why it needs to be dynamically loaded at runtime.

Captain-Calc:
The primary reason I left the textio_Input function behind was that I thought it was too rigid. It allowed the user to get input but made it difficult to for the user to create custom functions that manipulated the input (different keys for backspacing, moving the cursor, etc.). It also didn't cover other types of input, such as Hangman-like input or the input that the TI-BASIC Input function provides.

My goal with this version was to just provide functions that, as you said, merely manipulated the input buffer, leaving it up to the programmer how they wanted the input to be displayed. I also wanted to let them combine these simple routines with their own code to create functions that would manipulate the buffer.

But with that being said, I would be open to a ground-up redesign of the library. The TextIOC thread is still open on Cemetech, and I would be happy to discuss API designs there. We can also get the calculator community involved and see what they would like to have in the library. Tell me what you want, and I will see what I can do.

PR #274, opened 2020-08-15
Enable generation of MANPAGES on doxygen config

TheRealGlumfish:
Enable the generation of Unix Man pages. MAN PAGES are highly useful for quickly finding out information on functions and headers. The MAN PAGES generated by doxygen in the project's current configuration seem to be usable. Maybe some configuration could be done to add a better header. This could be also added to the wiki to enable the MAN PAGES to be seen by the `man` utility. 

> 3. (Optional) Append and export the `MANPATH` environment variable using the repository's `doxygen/man` folder 

> This can be done by adding this to the bash configuration file (.bash_rc or .bash_profile depending on the operating system) `export MANPATH=":/path/to/toolchain/doxygen/man/"`. Please use the appropriate path according to the location of the `toolchain` directory on your system.

### DIFF
--- a/tools/doxygen/config
+++ b/tools/doxygen/config
@@ -1862,7 +1862,7 @@ RTF_SOURCE_CODE        = NO
 # classes and files.
 # The default value is: NO.
 
-GENERATE_MAN           = NO
+GENERATE_MAN           = YES
 
 # The MAN_OUTPUT tag is used to specify where the man pages will be put. If a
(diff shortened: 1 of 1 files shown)


mateoconlechuga:
This is more of an issue than a pull request. There would need to be a fair amount of reasoning to this, as someone would have to maintain them and configure them properly. Someone can download and build them like you have; but it would be painful to manage since the primary base of the toolchain users are all on windows, with a sprinkling of mac.

PR #305, opened 2021-03-29
new keypadc v3 functions

beckadamtheinventor:
Added a few new functions for keypadc.
`uint8_t kb_ClearQueue(kb_queue_t *queue)` clears all keycodes from the keycode queue.
`uint8_t kb_QueueKeys(kb_queue_t *queue)` stores the keycodes of all the keys currently pressed. This should only be called if kb_UnqueueKey returns zero.
`uint8_t kb_UnqueueKey(kb_queue_t *queue)` recalls the lowest order keycode in the queue, removing it from the queue. This should be called repeatedly until it returns zero.
See `/examples/keypadc/key_queue/` for example usage of these functions.


runer112:
The "queue" is implemented as a bitset, and an inefficient one at that because it stores a full byte for each bit.

PR #323, opened 2021-08-09
added clarifications to fileioc

yolomep:
https://www.cemetech.net/forum/viewtopic.php?t=17915
I added the changes that I wanted. I just added one example function. Maybe next time someone who reads the documentation won't make the same mistake as me.


mateoconlechuga (commented):
accepted

Issue #321, opened 2021-07-13
Console Time


DuCouscous:
Is there any way with tice.h to get calc time in ms like System.currentTimeMillis(); in java ?

Thx ^^

Edit : The only one way i found is rtc_Time() but it's seconds, not ms :(

Edit #2 : Ik delay(ms) but this freeze the calc so it's not very noice

jacobly0:
The calculator hardware does not track real time to ms precision so no it is not possible.  If you just want a ms counter you can just use the timers in `tice.h`.

runer112:
This functionality is possible to provide if we set up one of the timers with this purpose in mind. I don't think we should discount this.

DuCouscous:
Thank you for help ^^

mateoconlechuga:
I disagree with Runer112. You would still need a function which starts the timer, and even then all you are making is very small wrappers around the already existing timer functions.

runer112:
The function to start the timer can be called from the startup code. And sure, it's only small wrappers, but that's kind of the point of system/standard libraries.

mateoconlechuga:
@runer112: There isn't a standard, non-implementation defined C function to get the system time in milliseconds though. Why not just let people use the already existing timer functions? There's no point in starting a random timer on program start.

runer112:
There isn't a standard C function, sure, but every OS provides one. And if we'd like to build up C++ support, then chrono needs it.

mateoconlechuga:
wut.

runer112:
I guess a better argument is for implementing https://en.cppreference.com/w/c/chrono/clock, which is part of the C standard library.

mateoconlechuga:
Yeah we should probably implement that rather than just returning -1. But what timer do we use? And what if the user starts to use it?

runer112:
Unless we want to only be able to count time for ~90 seconds, we'll have to use two timers, probably timers 1 and 2. If the program uses them for other purposes, there's really nothing we can do. `clock` (and any related functions) would simply produce garbage.

runer112:
@DuCouscous Good news, [clock()](https://en.cppreference.com/w/c/chrono/clock) has been implemented! In combination with [CLOCKS_PER_SEC](https://en.cppreference.com/w/c/chrono/CLOCKS_PER_SEC), you can now calculate time intervals with precision down to about a tenth of a millisecond.

PR #325, opened 2021-08-16
Implement clock() and sleep family of functions

runer112:
The clock uses a single 32768Hz 32-bit timer. It will overflow after ~36 hours, but that's probably acceptable for a calculator program.

Sleep functions added:
1. `msleep()`: millisecond sleep, "alias" of `delay()`
2. `sleep()`: second sleep, implements https://man7.org/linux/man-pages/man3/sleep.3.html
3. `ticksleep()`: clock tick sleep
4. `usleep()`: microsecond sleep, implements https://man7.org/linux/man-pages/man3/usleep.3.html


runer112:
Should address #321.

src/ce/crt0.src
```diff
@@ -107,6 +107,47 @@ launcher:
 .webaddress:
 	db	"http://tiny.cc/clibs",0
 
+
+	section .start
+
+mpTmrRange    := 0F20000h
+
+tmr2Counter   := 10h
```
runer112, line 115: Timer 2 is probably the best timer for this, right? I imagine that most programs with custom timer usage would use timer 1, and USB (including usbdrvce?) uses timer 3.

mateoconlechuga: That sounds good to me

examples/standalone/stopwatch/readme.md
```diff
@@ -1,7 +1,7 @@
 ### Stopwatch Demo
 
 This demo demonstrates using the hardware timers to make a simple stopwatch
-displaying millisecond precision.
+displaying hundreths precision.
```
adriweb, line 4: `hundredths`
Also "of seconds"?

src/makefile.mk
```diff
@@ -26,7 +26,7 @@ BSSHEAP_LOW ?= D052C6
 BSSHEAP_HIGH ?= D13FD8
 STACK_HIGH ?= D1A87E
 INIT_LOC ?= D1A87F
-OUTPUT_MAP ?= NO
+OUTPUT_MAP ?= YES
```
adriweb, line 29: was that done on purpose here

mateoconlechuga: I disagree with this modification

runer112: Yes, this was done on purpose. I don't see why anyone wouldn't want this extra build artifact that makes debugging in CEmu not a total nightmare.

Mateo mentioned something about it being slow on Windows, but I tried building oiram with and without this and saw no noticeable change in build time.

mateoconlechuga: Seems reasonable; I guess things have improved a lot

runer112:
Now that there are `sleep()` and `usleep()`, `delay()` to sleep for a number of milliseconds feels a bit out of place. How about introducing an alias, like `msleep()`?

mateoconlechuga:
> Now that there are `sleep()` and `usleep()`, `delay()` to sleep for a number of milliseconds feels a bit out of place. How about introducing an alias, like `msleep()`?

This makes sense to me.

examples/standalone/ticksleep/src/main.c
```diff
@@ -0,0 +1,16 @@
...
+int main(void)
+{
+    /* Clear the homescreen */
+    os_ClrHome();
+
+    /* Wait for 5 seconds */
+    os_PutStrFull("Waiting...");
+    ticksleep(5 * CLOCKS_PER_SEC);
+    os_PutStrFull(" Done!");
+    delay(100);
```
mateoconlechuga, line 13: why is there a delay here?

examples/standalone/usleep/src/main.c
```diff
@@ -0,0 +1,15 @@
...
+int main(void)
+{
+    /* Clear the homescreen */
+    os_ClrHome();
+
+    /* Wait for 5 seconds */
+    os_PutStrFull("Waiting...");
+    usleep(5 * 1000000);
+    os_PutStrFull(" Done!");
+    delay(100);
```
mateoconlechuga, line 12: and why a delay here... am I missing something?

src/ce/crt0.src
```diff
@@ -105,6 +105,47 @@ launcher:
...
+tmrCtrl       := 30h
+
+bTmr2Enable   := 3
+bTmr2Crystal  := 4
+bTmr2Overflow := 5
+bTmr2CountUp  := 10
+
+	public	___init_clock
```
mateoconlechuga, line 121: should this really be public? also it should be in a defined section

examples/standalone/ticksleep/src/main.c
runer112: That's just for the autotest to test the timing. It gives a 100ms target to check for after the expected time has elapsed.

src/ce/crt0.src
mateoconlechuga: is this code removed if the user isn't using clock() and related functions?

src/ce/delay.src
```diff
@@ -1,90 +1,64 @@
...
+; Get millis.
 	pop	de
-	ex	(sp),hl			; 0hl = millis
+	ex	(sp), hl		; hl = millis
 	push	de
+; ^ 48 cc
 ; Abort ASAP if millis == 0.
-	ld	a,h
-	or	a,l
+	ld	a, l
```
mateoconlechuga, line 20: this is unsafe

src/ce/sleep.src
```diff
@@ -0,0 +1,42 @@
...
+; Even though the input is 24-bit, let's just pretend that it's 16-bit.
+; Timing assumes fetching from RAM (4cc) and the first timer read being valid.
+
+; Get seconds.
+	pop	de
+	ex	(sp), hl		; hl = seconds
+	push	de
+; ^ 48 cc
+; Abort ASAP if seconds == 0.
+	ld	a, l
```
mateoconlechuga, line 21: this is technically unsafe

src/ce/crt0.src
runer112: Yes, it needs to be public. Referencing this symbol (or not) is what determined whether the section it's in (and it is in one) will be included.

src/ce/usleep.src
```diff
@@ -0,0 +1,60 @@
+; typedef unsigned int useconds_t;
+; int usleep(useconds_t usec);
+
+	assume	adl=1
+
+virtual
+	call	_usleep
```
mateoconlechuga, line 7: why?

mateoconlechuga (changes requested):

src/ce/crt0.src
mateoconlechuga: Oh that makes sense.

runer112: I just double-checked. Yes, if none of the new timing functions are referenced, this initialization code is not included.

mateoconlechuga (approved):

mateoconlechuga (approved):

mateoconlechuga (approved):

PR #328, opened 2021-08-27
Add os_RunPrgm [fix #317]

### LINKED ISSUE #317: Function to run another program
Might get y'all to stop complaining about everyone else's attempts at writing one :p

More seriously, there's obviously a demand for such a function given that there are multiple people writing their own implementations.

As for the interface, should there be two separate functions, one that exits after running the program and one that re-runs the current program? Also, should there be some mechanism to signal to the program that it's being re-run, or should the user be required to implement that themselves?

mateoconlechuga:
Calls assembly and basic programs. Uses a stub on the stack to reload, as well as saved user information. A marker word is used on the stack when a program is run to know whether to execute the callback or not. Applies to #317


src/ce/tice.h
```diff
@@ -919,6 +925,33 @@ int os_PushErrorHandler(void) __attribute__((returns_twice));
...
+ * will contain the error code if a TI-BASIC program, or the exit code if a C
+ * program. Other types of programs may have an undefined \p retval. This
+ * argument may be left \c NULL if execution should not return to the calling
+ * program.
+ *
+ * @return This function should not return, but if it does, -1 indicates the
+ * program could not be found, -2 if not enough memory, and < 0 if some other
+ * error occurred.
+ */
+int os_RunPrgm(const char *pgrm, void *data, size_t size, int (*ret)(void *data, int retval));
```
jacobly0, line 953: Why aren't you using the typedef here?
```suggestion
int os_RunPrgm(const char *pgrm, void *data, size_t size, os_runprgm_callback_t callback);
```

src/ce/crt0.src
```diff
@@ -143,21 +141,67 @@ assert mpTmr2Counter shr 8 = mpTmrCtrl shr 8
...
+	ld	hl,0
+.retval := $ - 3
+	push	hl
+	ld	hl,0
+.data := $ - 3
+	push	hl
+	call	0
+.callback := $ - 3
+	pop	de, de
+	ret
```
jacobly0, line 164: There's no reason to fix sp, main arguments are certainly never manually removed.
```suggestion
	ex	(sp),hl
	ld	de,0
.data := $ - 3
	push	de,hl
	jp	0
.callback := $ - 3
```

jacobly0 (changes requested):
All program return codes are super broken, noted `crt0.src` optimizations.

src/ce/crt0.src
```diff
@@ -438,19 +488,20 @@ end virtual
 _abort:
 	ld	hl,1 shl 23+6	; SIGABRT
 
-	section	.start
+	section	.fini
 	private	skip._abort
 skip._abort:
+	pop	hl
```
jacobly0, line 494: This doesn't make any sense once you fix the section bugs.
```suggestion
```

src/ce/crt0.src
```diff
@@ -143,21 +141,67 @@ assert mpTmr2Counter shr 8 = mpTmrCtrl shr 8
...
+	ld	(___rerun_handler.callback),de
+	pop	de
+	ld	(___rerun_handler.data_size),de
+	or	a,a
+	sbc	hl,hl
+	add	hl,sp			; user callback data
+	ld	(___rerun_handler.data),hl
+	jq	__start
+.no_return:
+	push	hl
```
jacobly0, line 191: Optimize better.
```suggestion
	ld	de,1
	add	hl,de
	jq	nc,.no_rerun
	ld	hl,___rerun_handler
	ld	(__start._main),hl	; doesn't work for custom start?
	pop	hl
	ld	(___rerun_handler.retval),hl
	pop	hl
	ld	(___rerun_handler.callback),hl
	pop	hl
	ld	(___rerun_handler.data_size),hl
	ex	de,hl			; hl = 1
	add	hl,sp			; user callback data
.no_rerun:
	dec	hl
	ld	(___rerun_handler.data),hl
	push	hl
```

jacobly0 (changes requested):
Fix more bugs revealed by previous bugfixes and another `crt0.src` optimization.

src/ce/crt0.src
```diff
@@ -143,21 +141,67 @@ assert mpTmr2Counter shr 8 = mpTmrCtrl shr 8
...
+.callback := $ - 3
+	pop	de, de
+	ret
 
-	section	.start
-	private	start
-start:
+	section .init
+	public	___prgm_run_check
+___prgm_run_check:
```
jacobly0, line 168: This should be after `save sp for exiting`, so that you can jump straight to the rerun handler and not need all the smc or need to run ans parsing code if main isn't going to be run anyway.  This is not a custom start issue because if it's your job to call main anyway then it's also your job to detect rerun and act accordingly.  However there is still a custom start issue because just putting it in the `.init` section does not guarantee it's location relative to all the other conditional `.init` code.

mateoconlechuga: It can't be directly after that because we still need to initialize the bss and constructors.

src/ce/tice.h
mateoconlechuga: fixed and updated

src/ce/crt0.src
mateoconlechuga: fixed and updated

src/ce/crt0.src
mateoconlechuga: fixed and updated

src/ce/crt0.src
```diff
@@ -321,26 +365,32 @@ else
...
 	require	skip._exit
-
-	section	.start
 virtual
 	ld	a,0
  load .ld_a: byte from $$
 end virtual
 	db	.ld_a
 
+	section	.fini
```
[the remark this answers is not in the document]
mateoconlechuga: fixed and updated

src/ce/crt0.src
```diff
@@ -321,26 +365,32 @@ else
...
 
-	call	main
+	section	.fini
 if push_exit_code <> pop_exit_code
 	push	hl
 end if
 push_exit_code:
 	require	skip._exit
-
-	section	.start
```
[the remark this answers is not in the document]
mateoconlechuga: fixed and updated

src/ce/crt0.src
```diff
@@ -321,26 +365,32 @@ else
...
+ifextern ___main_argc_argv
+	call	___main_argc_argv
+else
+	call	_main
+end if
+	public	__start._main
+__start._main := $ - 3
 
-	call	main
+	section	.fini
```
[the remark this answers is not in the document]
mateoconlechuga: fixed and updated

src/ce/crt0.src
```diff
@@ -143,21 +140,65 @@ assert mpTmr2Counter shr 8 = mpTmrCtrl shr 8
 	ld	l, mpTmrCtrl and $FF
 	set	bTmr2Enable, (hl)
 
+	section	.text
+	private	___rerun_handler
+	private	___rerun_handler.callback
+	private	___rerun_handler.data
+	private	___rerun_handler.retval
```
jacobly0, line 147:
```suggestion
	private	___rerun_handler
	private	___rerun_handler.callback
	private	___rerun_handler.data
	private	___rerun_handler.retval
	private	___rerun_handler.data_size
```

src/ce/crt0.src
```diff
@@ -143,21 +140,65 @@ assert mpTmr2Counter shr 8 = mpTmrCtrl shr 8
...
+	ld	(___rerun_handler.callback),de
+	pop	de
+	ld	(___rerun_handler.data_size),de
+	or	a,a
+	sbc	hl,hl
+	add	hl,sp			; user callback data
+	ld	(___rerun_handler.data),hl
+	jq	__start
+.no_return:
+	push	hl
```
jacobly0, line 188: `__start._main` should probably be a weak symbol accessible from C so that it can define its own to call indirectly and not have to have the correct declaration of main visible while also supporting rerun.

Also, actually working optimization:
```suggestion
	ld	de,1
	add	hl,de
	dec	hl
	jq	nc,.no_rerun
	ld	hl,___rerun_handler
	ld	(__start._main),hl	; doesn't work for custom start?
	pop	hl
	ld	(___rerun_handler.retval),hl
	pop	hl
	ld	(___rerun_handler.callback),hl
	pop	hl
	ld	(___rerun_handler.data_size),hl
	ex	de,hl
	add	hl,sp			; user callback data
	dec	hl
	ld	(___rerun_handler.data),hl
virtual
	ld	a,0
 load .ld_a: byte from $$
end virtual
	db	.ld_a
.no_rerun:
	push	hl
```

jacobly0 (commented):
A minor cleanup and optimization.

src/ce/crt0.src
```diff
@@ -438,19 +491,19 @@ end virtual
...
 skip._abort:
 	ld	sp,0
+	public	_exit.sp
 _exit.sp := $-3
-	pop	iy		; iy = flags
-	pop	af		; a = original flash wait states
-	ex	(sp),hl		; hl = flash wait state control port,
-				; save exit code
-	ld	(hl),a		; restore flash wait states
+	push	hl
```
jacobly0, line 500:
```suggestion
	public	___exithl
___exithl:
	push	hl
```

src/ce/runprgm.src
```diff
@@ -0,0 +1,400 @@
...
+	lddr 
+.done_move:
+	ld	hl,(_exit.sp)
+	ld	de,0			; fake return
+.fakeret := $-3
+	ld	(hl),de
+	or	a,a
+	sbc	hl,hl
+	push	hl
+	call	_exit			; should return to caller
```
jacobly0, line 179:
```suggestion
	jp	___exithl			; should return to caller
```

src/ce/tice.h
```diff
@@ -919,6 +925,33 @@ int os_PushErrorHandler(void) __attribute__((returns_twice));
...
+/**
+ * Runs a program that exists on the calculator.
+ * Note that this will destroy the currently running program, requiring you to
+ * save any data as needed. This program has an optional callback that will be
+ * executed when the called program finishes, which can be used to rebuild the
+ * program state. Additionally, program context information can be safely
+ * stored by using the extra user data arguments, which will then be delivered
+ * to the callback.
+ *
+ * @param pgrm Name of program to execute. The type should be the first byte of
```
jacobly0, line 937:
```suggestion
 * @param prgm Name of program to execute. The type should be the first byte of
```

src/ce/tice.h
```diff
@@ -919,6 +925,33 @@ int os_PushErrorHandler(void) __attribute__((returns_twice));
...
+ * will contain the error code if a TI-BASIC program, or the exit code if a C
+ * program. Other types of programs may have an undefined \p retval. This
+ * argument may be left \c NULL if execution should not return to the calling
+ * program.
+ *
+ * @return This function should not return, but if it does, -1 indicates the
+ * program could not be found, -2 if not enough memory, and < 0 if some other
+ * error occurred.
+ */
+int os_RunPrgm(const char *pgrm, void *data, size_t size, os_runprgm_callback_t callback);
```
jacobly0, line 953:
```suggestion
int os_RunPrgm(const char *prgm, void *data, size_t size, os_runprgm_callback_t callback);
```

jacobly0 (commented):
Minor optimization and doc update.

examples/standalone/runprgm/makefile
```diff
@@ -0,0 +1,16 @@
+# ----------------------------
+# Makefile Options
+# ----------------------------
+
+NAME = DEMO
+ICON = icon.png
+DESCRIPTION =
```
commandblockguy, line 7:
```suggestion
DESCRIPTION = "CE C Toolchain Demo"
```
For consistency with the other examples

examples/standalone/runprgm/readme.md
```diff
@@ -0,0 +1,3 @@
+### Run Program
+
+This demo shows how to use os_RunPrgm to run programs stored on the OS.
```
commandblockguy, line 3:
```suggestion
This demo shows how to use os_RunPrgm to run programs stored on the calculator.
```

src/ce/tice.h
```diff
@@ -919,6 +925,33 @@ int os_PushErrorHandler(void) __attribute__((returns_twice));
...
+ * Runs a program that exists on the calculator.
+ * Note that this will destroy the currently running program, requiring you to
+ * save any data as needed. This program has an optional callback that will be
+ * executed when the called program finishes, which can be used to rebuild the
+ * program state. Additionally, program context information can be safely
+ * stored by using the extra user data arguments, which will then be delivered
+ * to the callback.
+ *
+ * @param prgm Name of program to execute. The type should be the first byte of
+ * the string.
```
commandblockguy, line 938: I don't see a use case for including the type byte. Users wanting to run BASIC from an appvar seems like an incredibly rare occurrence, and in my opinion it would be better to add a function to fileioc for changing the type of a variable so that users can change the appvar to a program if they want to run code from it.

examples/standalone/runprgm/src/main.c
```diff
@@ -0,0 +1,41 @@
+#include <tice.h>
+#include <stdlib.h>
+#include <stdio.h>
+
+int prgm2_ret(void *data, int retval)
+{
+    printf("executed prgm 2.\n");
+    printf("retval: %d", retval);
+    (void)data;
```
commandblockguy, line 9: Would it be better to use `__attribute__((unused))` for this? Or are we avoiding C extensions in the example code?

src/ce/runprgm.src
```diff
@@ -0,0 +1,399 @@
...
+	sbc	hl,hl
+	ld	l,a
+	ld	(ti.cursorImage),hl
+.no_error:
+	ld	iy,ti.flags
+	res	ti.progExecuting,(iy + ti.newDispF)
+	res	ti.cmdExec,(iy + ti.cmdFlags)
+	res	ti.allowProgTokens,(iy + ti.newDispF)
+	res	ti.textInverse,(iy + ti.textFlags)
+	res	ti.onInterrupt,(iy + ti.onFlags)
```
commandblockguy, line 320: This causes issues when calling a program that uses RunPrgm from a BASIC program, e.g.
```
Asm(prgmDEMO
Disp 1
```
will give an `ERR: INVALID`, since `allowProgTokens` is reset.

src/ce/runprgm.src
```diff
@@ -0,0 +1,399 @@
...
+.normal:
+	ld	(ti.cursorImage),hl
+	ld	iy,ti.flags
+	call	ti.PopErrorHandler
+	xor	a,a
+	sbc	hl,hl
+	ld	(ti.cursorImage),hl
+.basic_c_smc := $-3
+	jr	.no_error
+.error_handler:
```
commandblockguy, line 309: Not sure what's wrong with this, since I've never actually used error handlers before, but it seems to continue to execute the program after an error is encountered and the program calling RunPrgm exits. For example,
```
0/0
1
```
gives an `ERR: DATA TYPE` on the 1.
It seems to give an error unless the error occurs on the last line of the program, like it does in the test.

src/ce/tice.h
```diff
@@ -919,6 +925,33 @@ int os_PushErrorHandler(void) __attribute__((returns_twice));
...
+ * @param data User data that will be available in the callback function.
+ * May be \c NULL.
+ * @param size Size of user data (keep this small, it is stored on the stack!)
+ * @param ret Callback function to run when program finishes executing. The
+ * argument \p data will contain the provided \p data contents, and \p retval
+ * will contain the error code if a TI-BASIC program, or the exit code if a C
+ * program. Other types of programs may have an undefined \p retval. This
+ * argument may be left \c NULL if execution should not return to the calling
+ * program.
```
commandblockguy, line 947: Might want to mention that the return value of the callback is the return code of the program to the shell that ran the original program that called RunPrgm, assuming that's actually the case.

commandblockguy (commented):
Putting this here since GitHub doesn't let you comment on binary files for some reason:
I'm not sure if I like the fact that the .8xp files being run are included in the repository. On the one hand, it makes it easier to build the demo, and the programs are nearly empty. On the other, people who haven't used Git much before will be looking at these and we don't want to set an example of including binaries in a repo. Additionally, it means you can't tell what result the test is supposed to have without using a calculator, a hex editor, or convbin.

examples/standalone/runprgm/src/main.c
adriweb: Eh, we see this very often, looks fine to me.

src/ce/tice.h
jacobly0: I would say evaluating a function, but I imagine that is better served by another routine since functions can't call asm programs so you don't need to tear everything down.

mateoconlechuga: I disagree.

mateoconlechuga: Now I agree.

src/ce/runprgm.src
mateoconlechuga: This has been fixed by saving/restoring the state of the flags upon exectuion

src/ce/runprgm.src
mateoconlechuga: Fixed by saving the parser state before executing the basic program

jacobly0 (approved):

PR #335, opened 2021-09-11
Fix fileioc

### LINKED ISSUE #334: fileioc issues with old programs using new libs
So I noticed that the newest release of HexaEdit (as of 1 hour ago) works fine with the old libraries that were the latest at the time of its release, but for some reason it has issues with fileioc when using the new clibs.

runer112:
Fixes #334.


src/fileioc/fileioc.asm
```diff
@@ -1549,15 +1545,18 @@ util_is_slot_open:
 ;  c = slot
 ; out:
 ;  a = 0
+;  ubc = slot * 3
+;  uhl = pointer to upper byte of slot offset
 ;  zf = open
-	push	hl, bc
+;  (curr_slot) = slot
+	ld	a, c
+	ld	(curr_slot), a
```
runer112, line 1553: Having "optimized" this out was half of the issue.

src/fileioc/fileioc.asm
```diff
@@ -202,15 +202,15 @@ ti_Resize:
...
-	pop	de
-	pop	hl			; hl = new size
+	pop	hl
+	pop	de			; de = new size
 	pop	bc			; c = slot
 	push	bc
-	push	hl
 	push	de
+	push	hl
 	call	util_is_slot_open
```
runer112, line 211: `hl` being destroyed here was the other half of the issue. Now the new size is popped into `de` instead, which is not destroyed.

mateoconlechuga (approved):

PR #340, opened 2021-09-26
Build documentation in a few formats

tari:
This makes the documentation build script emit single-file HTML and LaTeX output as well as multi-file HTML since some users may prefer to have a single-file version of the documentation. The PDF currently doesn't build because it tries to use SVG graphics, however.

The documentation building action now also saves the documentation as an artifact.


jacobly0 (approved):

mateoconlechuga:
Thanks! I'll get the documentation to build shortly, and perhaps add a link to the offline docs in the webpage itself.

PR #345, opened 2021-12-20
Clarify os_RealCompare documentation.

Dragon-Hatcher:
This previously didn't specify when -1, 0, and 1 were returned. Additional research was required to determine when each was returned.


mateoconlechuga:
Thanks!

PR #355, opened 2022-01-27
Fix typos in fileioc.h

PeterTillema:


adriweb:
Well I guess it's better without the typo, but in theory it should be `curr_search_posistion` anyway :P

PR #384, opened 2022-05-20
Optimize popcount implementations

calc84maniac:
I've come up with a novel algorithm for more efficiently doing popcount on eZ80 (which really shines when operating on 16 bits at a time).

16-bit, 24-bit, and 32-bit implementations call into a common set of code for popcounting 16 bits efficiently. Meanwhile, the 8-bit implementation is also smaller and faster than the previous implementation. (Note that the previous 8-bit implementation would be more optimal in a Z80 CRT, though).

I also implemented unrolled, self-contained versions for 16-bit, 24-bit, and 32-bit, which may be useful once the toolchain supports linking them.

I exhaustively tested the 8-bit and 16-bit implementations, and tested the 24-bit and 32-bit implementations within reasonable doubt.

These are my calculations for the cycle savings compared to the previous impls (assuming standard CE RAM waitstates):
- bpopcnt: -12.2%
- spopcnt: -43.6%
- ipopcnt: -14.8%
- lpopcnt: -25.2%

As for the sizes, the only standalone routine (bpopcnt) has been reduced by 5 bytes (-20%).
The others are more complicated due to the shared code:
- If only spopcnt is used, the new implementation is 5 bytes smaller (-12.5%).
- If only ipopcnt is used, the new implementation is 1 byte larger (+2.9%).
- If only lpopcnt is used, the new implementation is 12 bytes larger (+42.9%).
- If all routines are used, the new implementation is 14 bytes larger in total (+18.7%).

### DIFF
--- a/src/crt/bpopcnt.src
+++ b/src/crt/bpopcnt.src
@@ -3,25 +3,30 @@
 	section	.text
 	public	__bpopcnt
 __bpopcnt:
-	push	bc
-	ld	c,a		; c=(A|B|C|D|E|F|G|H)
-	and	a,$aa		; a=(A|0|C|0|E|0|G|0)
-	cpl			; a=(~A|1|~C|1|~E|1|~G|1)
-	rrca			; a=(1|~A|1|~C|1|~E|1|~G), cf=1
-	adc	a,c		; a=(A+B|C+D|E+F|G+H)
-	ld	b,a		; b=(A+B|C+D|E+F|G+H)
-	and	a,$33		; a=(00|C+D|00|G+H)
-	ld	c,a		; c=(00|C+D|00|G+H)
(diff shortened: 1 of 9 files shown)


jacobly0:
Tests pass, saves 20/165/51/103/206 cycles.

jacobly0 (approved):

runer112 (commented):
I'd like to look at this a bit more closely when it's not 11:58pm on a weekday, but at a glance and having understood the approach, I like it.

mateoconlechuga (approved):

jacobly0:
All tests continue to pass.

runer112 (approved):

mateoconlechuga:
Thanks for the PR! Merged :)

PR #400, opened 2022-08-11
Optimize 24-bit right shift for much lower maximum runtime

calc84maniac:
24-bit right shifts up until now have had linear time complexity on the number of shifts, which balloons to an insane amount of cycles for any non-trivial number of shifts.

This proposed update limits the maximum runtime for any shift amount to just below the time it previously took to shift by 2, at the cost of a 50% increase in code size. Unfortunately, shifts by 1 are about 30% slower, but ideally the compiler should be using the __ishr*_1_fast routines for that instead for constant shifts. Also, shifts by 0 no longer have an early-out path, but I feel that has very limited benefit in the context of variable shift amounts.


jacobly0:
```
Testing __ishrs... 0.000000% failed in 38/386.737775/1742 cycles minimum/average/maximum
Testing __ishru... 0.000000% failed in 38/396.233433/1759 cycles minimum/average/maximum
```
to
```
Testing __ishru... 0.000000% failed in 195/254.181556/267 cycles minimum/average/maximum
Testing __ishrs... 0.000000% failed in 175/234.181556/247 cycles minimum/average/maximum
```

jacobly0 (approved):

runer112 (approved):
Very nice!

mateoconlechuga (approved):

adriweb:
Merged (the failed build wasn't related - bypassed)

PR #402, opened 2022-08-12
Optimize __ishl for bounded execution time for larger shifts

calc84maniac:
Similarly to the recent right-shift update, this bounds the execution time of 24-bit left shifts. It's slower for shifts by 0, 1, or 2, but the same or much faster for all other shifts. Constant shifts by 0-4 should be inlined by the compiler to avoid calls to this routine in the first place.

### DIFF
--- a/src/libc/ishl.src
+++ b/src/libc/ishl.src
@@ -10,18 +10,38 @@ __ishl := 000174h
 else
 
 __ishl:
-; Suboptimal for large shift amounts
-; CC: if C!=0: C*(4*r(PC)+1)+7*r(PC)+(ADL?6*r(SPL)+3*w(SPL):4*r(SPS)+2*w(SPS))+1
-;     if C==0: 4*r(PC)+(ADL?3*r(SPL):2*r(SPS))+2
-	inc	c
-	dec	c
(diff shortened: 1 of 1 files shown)


src/libc/ishl.src
```diff
@@ -10,18 +10,38 @@ __ishl := 000174h
...
+	ret	z
+__ishl_16:
+	ld	h, l
+	ld	l, 0
+__ishl_8:
+	push	hl
+	dec	sp
+	pop	hl
+	inc	sp
+	ld	l, 0
```
runer112, line 44: This implementation seems to be largely optimized for speed, in which case, this should probably be 8x `add hl, hl`.

calc84maniac: I was trying to keep it somewhat balanced between size and speed, but what's another 2 bytes in the grand scheme of things?

jacobly0 (approved):
```
Testing __ishl... 0.000000% failed in 38/128.203411/458 cycles minimum/average/maximum
```
=>
```
Testing __ishl... 0.000000% failed in 116/128.246863/179 cycles minimum/average/maximum
```

PR #404, opened 2022-08-16
Optimize and/or/xor/ldivs, combine functions that reference each other

calc84maniac:
I found an optimization for 24-bit and 32-bit and/or/xor operations which saves 2 bytes/fetches per routine. In the process, I also realized that 24-bit and 32-bit could share implementations to save space.

But, I didn't want to force the 32-bit implementations to be included in programs that didn't use them, so @jacobly0 taught me some linker magic I could use for conditional inclusion of code blocks in the same file. I went and applied this magic to each set of CRT functions which shared an implementation, to reduce JPs to JRs in some cases, or to prevent unused code inclusion in other cases.

Also, in the process of rearranging functions, I implemented a minor optimization to __ldivs for checking opposite signs.


src/crt/idvrmu_sdvrmu.src
```diff
@@ -1,5 +1,29 @@
...
+	ld	e, 0
+
+	inc	bc
+	dec.s	bc
+
+	push	af
+	ld	a, 16
+
+if PREFER_OS_CRT
+	require	__idvrmu_common
```
runer112, line 22: It looks like this was done improperly by me to begin with, but if `PREFER_OS_CRT` is set, I think we should probably just zero-extend `HL` and `BC` and then jump to `__idvrmu`.

src/crt/idvrmu_sdvrmu.src
```diff
@@ -1,5 +1,29 @@
 	assume	adl=1
 
+	section	.text
+	public	__sdvrmu
+__sdvrmu:
+; I: HL=dividend, BC=divisor
+; O: a=0, ude=HL/BC, uhl=HL%BC
```
runer112, line 7: Another error of mine:

```suggestion
; O: ude=HL/BC, uhl=HL%BC
```

runer112:
Is there perhaps a better naming strategy we can use for files that define multiple symbols? Just joining them with an underscore is pretty confusing.

src/libc/ishr.src
```diff
@@ -1,12 +1,10 @@
 	assume	adl=1
```
runer112, line 1: I just noticed: What is this doing in _libc_ instead of _crt_ like everything else?

src/crt/lneg_labs_ldivs_lrems.src
```diff
@@ -0,0 +1,109 @@
...
+; O: euhl=EUHL/AUBC
+
+	rlc	e
+	rrc	e
+	rra
+	adc	a, a
+	push	af
+	push	bc
+
+	call	__ldivs_lrems_common
```
runer112, line 65: I believe that this optimization is invalid:
https://github.com/CE-Programming/toolchain/blob/6c31e81cf6fd67bcae3be65e03e0abf619620f0a/src/crt/ldivs_lrems_common.src#L5-L6

runer112 (changes requested):

mateoconlechuga:
Yeah I'm not sure how I feel about multiple routines all in the same file - is there any way to do a 'include' instead or something?

mateoconlechuga:
What is the status of this PR? Perhaps some changes could be made into individual PRs that are easier to review (like one PR per function?)

calc84maniac:
Honestly, I kind of forgot about this after losing the hard drive I developed it on. I also forgot how ridiculously extensive it was, beyond the bitwise operator optimizations. I definitely feel like at least splitting it into families of functions would be a good idea.

mateoconlechuga:
All right I'll close it for now but feel free to open a new PR with changes :)

PR #418, opened 2022-12-05
Fix issues with ti_ArchiveHasRoom

mateoconlechuga:
This fixes issues with `ti_ArchiveHasRoom` (and subsequently `ti_SetArchiveStatus`). A new function is added to check if an open variable handle can be archived without a Garbage Collect as well (`ti_ArchiveHasRoomVar`).


src/fileioc/fileioc.asm
```diff
@@ -1621,19 +1643,28 @@ util_set_offset:
...
-	ex	hl,de
-	call	util_ArchiveHasRoom
-	jp	nz,ti.Arc_Unarc		; gc will not be triggered
+	ret	nz
+	call	ti_ArchiveHasRoomVar.entry_sym
+	jp	nz,ti.Arc_Unarc
+	call	ti.PushOP1
 	call	util_pre_gc_default_handler
 util_pre_gc_handler := $-3
 	call	ti.Arc_Unarc
```
calc84maniac, line 1656: Should there be a pop/push OP1 before Arc_Unarc, in case the pre-GC handler trashes it for some reason?

src/fileioc/fileioc.asm
```diff
@@ -1379,21 +1362,60 @@ ti_RclVar:
...
+	add	hl,bc
+	ld	c,(hl)			; get var size
+	call	util_get_data_ptr
+	ld	hl,(hl)
+.entry_sym:
+	ld	a,c
+	ex	de,hl
+	ld	hl,(hl)
 	ld	bc,12
 	add	hl,bc
```
calc84maniac, line 1408: What if this addition carries? I recommend doing add a,c \ ld c,a instead of add hl,bc \ ld c,a. Name size plus 12 is guaranteed to fit in 8 bits, and then we get the combined carry check on the 16-bit addition.

src/fileioc/fileioc.asm
```diff
@@ -1379,21 +1362,60 @@ ti_RclVar:
...
+	push	de
+	call	util_is_slot_open
+	jr	nz,.fail
+	call	util_get_vat_ptr
+	ld	hl,(hl)
+	ld	bc,-6
+	add	hl,bc
+	ld	c,(hl)			; get var size
+	call	util_get_data_ptr
+	ld	hl,(hl)
```
calc84maniac, line 1402: This should be ld de,(hl) to match the chkfindsym outputs as used below

src/fileioc/fileioc.asm
```diff
@@ -1379,21 +1362,60 @@ ti_RclVar:
 
 ;-------------------------------------------------------------------------------
 ti_ArchiveHasRoom:
```
calc84maniac, line 1364: If we're leaving the input as 24 bits, we should _probably_ do a range check (which effectively checks 16-bit carry for the caller). Also, .entry appears to be unused, so we're not constrained on implementation. I suggest:
```
ti_ArchiveHasRoom:
	pop	de
	pop	bc
	push	bc
	push	de
	cp	a, a	; Set Z flag
	ld	hl, $FF0000
	add	hl, bc
	call	nc, ti.FindFreeArcSpot
	ld	a, 1
	ret	nz
	xor	a, a
	ret
```

src/fileioc/fileioc.asm
mateoconlechuga: yep

PR #420, opened 2022-12-08
Default binaries to archive instead of RAM

LogicalJoe:


mateoconlechuga:
I'm not sure how I feel about this

adriweb:
I was actually the one proposing this on Discord due to talking about RAM issues.

The rationale behind it is that users generally don't mess with their transfer configuration on TI-Connect CE (nor do they archive stuff once it's on calc), so might as well have programs with sane defaults.

This might have been an issue a few years ago when 5.3 just came out, but now that it's been several years... it's fine.

(Also, the PB has been doing this for some time now :P)

mateoconlechuga:
If we do this, it might be worth removing the default from all of the examples too

adriweb:
> If we do this, it might be worth removing the default from all of the examples too

Yeah. This flag is already documented in the makefile options anyway, so if something needs to be in RAM, it will still be possible.

PR #428, opened 2023-02-08
includes: add names to some structs.

adriweb:
Became "necessary" for the lua binding thing that didn't like anonymous structs.

I'm not sure which naming convention we should use, though, and maybe `_` as prefix isn't great, but it's been used in some places before already so I used that too.
We could also just use `the_type_name` for the struct name, and the typedef would still be `the_type_name_t`?

### DIFF
--- a/src/ce/include/ti/real.h
+++ b/src/ce/include/ti/real.h
@@ -22,11 +22,11 @@ extern "C" {
 /**
  * @brief Structure of real variable type
  */
-typedef struct { int8_t sign, exp; uint8_t mant[7]; } real_t;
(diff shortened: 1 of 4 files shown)


adriweb:
Alright. 

I don't know what's up with the CI though, it builds fine for me and I'm not sure it's related since it's fasmg complaining. Can someone else check just in case?

runer112:
I'm most familiar with the convention `typedef struct foo foo_t`. If we didn't want that, I'd probably at least prefer `typedef struct foo_t foo_t`.

adriweb:
Oh I forgot that it was even possible to have both the struct name and the typedef be the same. Seems like it would be even better then?

PR #436, opened 2023-05-11
Add comments for symbols to grahpx.asm

TIny-Hacker:
I added comments for the symbols in the graphx font, to replace the `.` placeholder for symbols previously.


mateoconlechuga:
Haha nice - but not sure how I feel about this, doesn't this require the file to be utf8 compatible now?

adriweb:
I guess fasmg doesn't care, so it should be fine :)

adriweb:
So, hm, you put the corresponding symbols from the ASCII table? Because for the comments it would be better to just show the most closest-looking symbol in UTF-8.

TIny-Hacker:
I actually did try to find the closest looking symbol (As opposed to corresponding symbols from the ASCII table), it could be I made some mistakes though.

runer112 (commented):
I approve of this, assuming that this shouldn't cause any encoding issues. But someone else should probably confirm this.

TIny-Hacker:
I also fixed spacing for character "0x0D" (♪), which should've been 8 pixels instead of 2.

runer112 (approved):

PR #447, opened 2023-09-23
Change memset_fast macro to function definition

myclevorname:
[description]
On TI-Planet Project Builder, this does not show up as a function definition, shown below. It bothered me, so I decided to change it into a function definition.


myclevorname:
This probably was a stupid idea given *one* autocompletion popup.

PR #449, opened 2023-09-26
Optimize 24-bit and 32-bit and/or/xor operations

calc84maniac:
24-bit operations have been optimized to reduce two bytes/fetches. 32-bit operations are the same speed and size as before, but now save 23 bytes each when the 24-bit operation is also used.

For the "fast" variants, the 32-bit operations also receive the speed and size savings, and also save 7 bytes each when the 16-bit operation is used.

(Yes, I combined the files again, but I think the scope here is fine and the new filenames are appropriate. I recommend reviewing the individual commits.)


mateoconlechuga:
I like the combining of functions into one file :)

runer112 (approved):

PR #453, opened 2023-11-07
Added 48 bit Routines

fundudeone:
Added the following 48 bit crt routines:

slland: 348 cycles
sllor: 348 cycles
sllxor: 381 cycles
sllnot: 158 cycles
sllcmpu: 210 cycles
sllcmpzero: 45 cycles
sllshl: varies. It's quite long so it could be rewritten with a loop like shr routines if need be.
sllshru: varies
sllshrs: varies
sllmulu: 1592 cycles


src/crt/sllnot.src
```diff
@@ -0,0 +1,76 @@
...
+
+	; -6: ~l               <-- SP
+	; -5: ~h
+	; -4: ~u[hl]
+	; -3: e
+	; -2: d
+	; -1: u[de]
+	; -0: everything else
+	
+	inc sp
```
calc84maniac, line 32: This technique is invalid if interrupts are enabled. Anything below (SP) can be considered trashed at any time.

fundudeone: Revised version no longer uses this technique

src/crt/sllmulu.src
```diff
@@ -0,0 +1,280 @@
...
+; 
+; Returns:
+; ude:uhl = ude:uhl * uiy:ubc
+
+	assume	adl=1
+
+	section	.text
+	public	__sllmulu
+__sllmulu:
+	;backup af, alt bc, alt de, and alt hl
```
calc84maniac, line 17: The issue with using shadow registers isn't that they need to be backed up, but that they can be clobbered by interrupts. For a long routine like this, it may be worth disabling interrupts and restoring them at the end (32-bit and 64-bit division in the toolchain already do this).

src/crt/sllshl.src
```diff
@@ -0,0 +1,125 @@
...
+	add	iy, iy
+	adc hl, hl
+	add	iy, iy
+	adc hl, hl
+	add	iy, iy
+	adc hl, hl
+.bit2:
+	bit	3, c
+	jr	z, .bit3
+	add	iy, iy
```
calc84maniac, line 48: Unrolling a left shift by 8 like this may be fast if it's just HL, but across two registers like this, it's slower (and far more bloated) than the stack-based equivalent. The same goes for the other byte-aligned shifts below as well.

src/crt/sllshl.src
```diff
@@ -0,0 +1,125 @@
...
+
+assume	adl=1
+
+	section	.text
+	public	__sllshl
+__sllshl:
+	push iy
+	push de
+	push hl
+	pop iy ; iy will be used for least significant, hl for most going forward
```
calc84maniac, line 20: I think it would be less overhead overall if you don't juggle the registers, but instead do DE/HL swaps in the shifts, like:
```
add hl, hl
ex de, hl
adc hl, hl
ex de, hl
```
This is technically 1 more byte per single shift, but there should only be 7 single shifts total in the code (if the byte-aligned shifts are implemented via the stack), which is less overhead than the register juggling.

src/crt/sllshrs.src
```diff
@@ -0,0 +1,49 @@
...
+
+
+; Returns:
+; ude:uhl = ude:uhl>>c
+
+	assume	adl=1
+
+	section	.text
+	public	__sllshrs
+__sllshrs:
```
calc84maniac, line 16: This needs an early return if shifting by 0, otherwise it incorrectly shifts by 256.

src/crt/sllshrs.src
```diff
@@ -0,0 +1,49 @@
...
+	dec hl
+	rr (hl)
+	dec hl
+	rr (hl)
+	dec hl
+	rr (hl)
+	dec hl
+	rr (hl)
+
+	or a, a
```
calc84maniac, line 43: This carry clear seems irrelevant.

src/crt/sllshrs.src
```diff
@@ -0,0 +1,49 @@
...
+	assume	adl=1
+
+	section	.text
+	public	__sllshrs
+__sllshrs:
+	push bc
+	ld b, c
+	push de
+	push hl
+	ld hl, 0
```
calc84maniac, line 21: You could make this a `ld hl,5 \ add hl,sp` and move it into the loop to save space, since that has the same cost as 5 increments.

src/crt/sllshru.src
```diff
@@ -0,0 +1,51 @@
...
+	push de
+	push hl
+	ld hl, 0
+	add hl, sp
+	
+	inc hl
+	inc hl
+	inc hl
+	inc hl
+	inc hl
```
calc84maniac, line 28: This all can also be a `ld hl, 5 \ add hl, sp`

src/crt/sllshru.src
```diff
@@ -0,0 +1,51 @@
...
+	
+	srl (hl)
```
calc84maniac, line 30: This is up to you, but if you want to save a bit of space you could jump straight into the middle of the signed loop from here.

src/crt/sllcmpzero.src
```diff
@@ -0,0 +1,28 @@
...
+
+	section	.text
+	public	__sllcmpzero
+__sllcmpzero:
+	;check lsb's == 0
+	add hl, bc
+	or a, a
+	sbc hl, bc
+	ret nz
+	;check msb's == 0
```
calc84maniac, line 21: Since you know at this point that UHL==0, you could simplify this part into an UHL==UDE comparison

src/crt/sllcmpu.src
```diff
@@ -0,0 +1,70 @@
...
+	assume	adl=1
+
+	section	.text
+	public	__sllcmpu
+__sllcmpu:
+	push ix
+	ld ix, 0
+	add ix, sp
+
+	ld hl, (ix+18) ;hl = iy
```
calc84maniac, line 21: Are the parameters passed on the stack for this function? If so, update the documentation

src/crt/slland.src
```diff
@@ -0,0 +1,68 @@
...
+	assume	adl=1
+
+	section	.text
+	public	__slland
+__slland:
+	push af
+
+	; Set up stack  stuff for later
+
+	push ix
```
calc84maniac, line 21: I recommend using IY as a frame pointer after you push it later on, then you don't have to save and restore IX

src/crt/sllmulu.src
fundudeone: Fixed by Updated sllmulu

calc84maniac: Sorry, maybe I should have clarified, but when I say "restoring" interrupts, that means checking whether they're enabled or not before disabling them, and then conditionally enabling them at the end. We definitely don't want a multiplication to randomly enable interrupts, after all.

src/crt/sllcmpu.src
fundudeone: Values aren't passed onto the stack during the routine, but to be slightly faster, they are taken off the stack because I saw llcmpu do something similar. Documentation is updated.

calc84maniac: llcmpu could probably use some documentation too, but it seems to be passing one 64-bit parameter in registers and the other on the stack (since there aren't enough registers for both).

runer112:
Before I really get into review, just checking, does the calling convention for these routines match what ez80-clang emits?

fundudeone:
> Before I really get into review, just checking, does the calling convention for these routines match what ez80-clang emits?

Yes, they should work with the Z80_LibCall calling convention (once everything's fixed at least)

mateoconlechuga:
I don't like the fact that some CRT routines are disabling interrupts... but we can't really use them anyway so maybe I'm less conflcted :upside_down_face:

runer112 (commented):
I only reviewed a small portion of the changes so far. Before I continue, we really ought to get @jacobly0's support to (a) confirm the signature expected for each of these functions and (b) produce a version of ez80-clang that actually emits calls to dedicated 48-bit math functions so they can be tested.

calc84maniac:
Just FYI, support has now been added to the compiler for these routines, and they use an `i48` naming scheme rather than `sll`. I've added stub implementations for these in [this file](https://github.com/CE-Programming/toolchain/blob/d94377a4173fce8790834b24fc4d9041df2f0adb/src/crt/i48stubs.src) which ideally should make the routines "work" (suboptimally) by wrapping the 64-bit routines, and those can be replaced with full implementations via this PR.

My observations about the libcall calling convention (hopefully accurate, haven't confirmed but should be testable now) are:
Routines that take one 48-bit parameter: input in UDEUHL
Routines that take two 48-bit parameters: inputs in UDEUHL, UIYUBC
Routines that take one 48-bit parameter and one 8-bit parameter: inputs in UDEUHL, C
Routines that return one 48-bit result: output in UDEUHL
Routines that return two 48-bit results (just `__i48dvrmu`): output in UDEUHL, UIYUBC
Routines that return one 8-bit result: output in A
`__i48cmpu` outputs C and Z flags
`__i48cmpzero` outputs S and Z flags
`__i48cmps` outputs S and V flags (does it also need the Z flag? the 32-bit and 64-bit implementations seem to indicate no, but they could be buggy if it's needed)

src/crt/slland.src
```diff
@@ -0,0 +1,63 @@
...
+
+	ld a, (iy-4) ; a = u[hl] (modified)
+	and a, (iy-7) ; a &= u[bc]
+	ld (iy-4), a ; u[hl] (modified) = a
+
+	; Load values back into registers and clean up stack
+	pop bc
+	pop hl
+	pop de
+	pop iy
```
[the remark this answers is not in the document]
fundudeone: Sorry this took me so long to get back to. This makes sense and works with everything I've thrown at it, so go ahead with or and xor.

adriweb:
Any update? @runer112 or @calc84maniac maybe?
(there's one thing that needs to be "resolved" for the test to start apparently)

calc84maniac:
Just rebased on top of my newly added regression test for these routines, it's identified a few issues in this PR so I'm going to work on fixing those, and then look at implementing optimizations which the test will also help verify.

calc84maniac:
Okay, I'm done with my full bugfix and optimization pass. If @runer112 or any other reviewers want to take a look at it now, feel free.

runer112 (approved):
If @calc84maniac has already had a good look at each of these routines, that's satisfactory to me.

PR #460, opened 2023-11-19
Experimental support for CH340, FT232RL & CP2101 devices

tr1p1ea:
Experimental support for CH340, FT232RL & CP2102 devices.

Tested with ESP8266 D1 mini clone, ESP-WROOM-32 board and FTDI232 boards.

Currently baud rates are table driven so baud rates are limited to: 9600, 19200, 38400, 57600 & 115200 but can be easily expanded.


mateoconlechuga:
Looks good to me!

mateoconlechuga:
Does the C header need updating?

tr1p1ea:
I fixed an issue with CH340 clones and I'm adding CH341 today.

I'll also update the header file.

adriweb:
rebased and launched CI just to check

src/srldrvce/srldrvce.h
```diff
@@ -77,10 +110,14 @@ typedef struct {
...
     srl_device_type_t type;
     srl_device_subtype_t subtype;
+    uint16_t idvendor;
+    uint16_t idproduct;
+    uint16_t bcddevice;
+    uint8_t baudtype;
     ring_buf_ctrl_t rx_buf;
     ring_buf_ctrl_t tx_buf;
     srl_error_t err;
-    uint8_t reserved[16];
```
mateoconlechuga, line 83: I'm a little worried about this change - does this break out backwards compatibility when user allocated structures might be broken? @commandblockguy 

The issue is someone using a program compiled for an old version of the lib, and using a newer lib.

adriweb: Could new fields be added at the end instead, and the reserved array shortened?

mateoconlechuga: This is why the structure should be big enough to support forwards compatibility and the addition of fields - also moving of fields breaks backwards compatibility for all programs too.

mateoconlechuga: @adriweb no, because the structure is user-allocated. so using a newer library would corrupt memory.

commandblockguy: Memory corruption could only happen if the total size of the struct changes, right?
In hindsight, yeah, it probably wasn't a good idea to make the internal structures visible - I mostly only did that to ease debugging, and didn't get around to removing them before actually releasing it.
Aside from the device field, I don't imagine that people have used the struct fields - it would probably be a good idea to deprecate them.

mateoconlechuga:
@tr1p1ea what is the status of this PR? Also @commandblockguy, which struct fields can we deprecate to make everything reserved?

commandblockguy:
We can probably deprecate everything except for the USB device handle.

mateoconlechuga:
Great :)

tr1p1ea:
After testing I've come to the conclusion that there is some further work to do to improve compatibility with some devices.

Should I recall this PR until that's complete?

Also did you want the struct optimisation to be part of this too?

mateoconlechuga:
@tr1p1ea yes to the struct optimization and any fun new status updates for us :)

PR #468, opened 2024-01-06
Fix heading spelling of keypadc.h docs

myclevorname:
Dection → Detection

Pretty self-explanatory.

### DIFF
--- a/docs/libraries/keypadc.rst
+++ b/docs/libraries/keypadc.rst
@@ -13,13 +13,13 @@ It supports multi-key presses, and can be used for extremely responsive input co
 .. contents:: :local:
    :depth: 3
 
-Helpful key detection
+Key Detection
 -------------------
 
 Handling key detection properly is a major requirement for a lot of programs.
 Detecting a hold can be rather straightforward, but things like a press, release, can sometimes be rather confusing at first.
 
-Detecting a single press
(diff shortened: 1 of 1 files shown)


docs/libraries/keypadc.rst
```diff
@@ -13,7 +13,7 @@ It supports multi-key presses, and can be used for extremely responsive input co
 .. contents:: :local:
    :depth: 3
 
-Helpful key dection
+Helpful key detection
```
runer112, line 16: I just noticed, why is the word "Helpful" here? Is it just me thinking that it sounds weird with it?

myclevorname: I will change that too.

runer112 (approved):

myclevorname:
Maybe it could be 'Detecting Keys' instead of 'Key Detection'. What are your thoughts?

runer112:
> Maybe it could be 'Detecting Keys' instead of 'Key Detection'. What are your thoughts?

I think what you already have with "Key Detection" is good.

runer112:
I'll wait a bit for any other potential reviewers to get a chance, but short of any other reviews, I'm happy to merge this. Seems like a simple and objective improvement.

PR #474, opened 2024-01-29
More GetStringInput size optimizations

calc84maniac:
Builds off of the previous PR to implement more size optimizations (including some JP to JR conversions due to the reduced size, which is why I'm making a PR to verify the build)


mateoconlechuga:
@calc84maniac is this good to be merged?

calc84maniac:
Yes, as far as I know, but it should be squashed I think

PR #480, opened 2024-02-27
strncasecmp fix and optimizations

### LINKED ISSUE #479: Incorrect behavior in strncasecmp and TI-OS implementations of tolower/toupper
The [current implementation](https://github.com/CE-Programming/toolchain/blob/30cfdf28d1391bf36a98c5aedaab3427c823916e/src/libc/strncasecmp.src) of `strncasecmp` appears to be calling `tolower` with values outside of the `unsigned char` range, which is UB and (in the toolchain implementation) returns values which are unsuitable for 24-bit integer comparison. In the TI-OS implementation of `tolower`, it effectively treats the input as `unsigned char` , which possibly hid this issue in `strncasecmp`. However, that means the TI-OS implementation is not standard-compliant because it doesn't return EOF when given EOF as input.

In short, we can probably ditch the TI-OS implementations of `tolower`/`toupper`, and possibly provide an internal ASM entrypoint in our own `tolower` to make the `strncasecmp` implementation much more efficient and clean (and more importantly, correct).

Edit: Thanks to @envenomator for reporting this bug on Discord.

calc84maniac:
Fixes #479 with a new `strncasecmp` implementation.

Removes linking to OS versions of `tolower`/`toupper` which return the wrong value for EOF, and adds internal entrypoints to the toolchain versions to avoid C ABI overhead.

Also adds an optimized version of `strcasecmp`, which shares some code with the new `strncasecmp` for size optimization.


src/libc/tolower.src
```diff
@@ -3,24 +3,20 @@
...
+	public	_tolower.internal
+	; ASM interface: input/output char in L, destroys AF
+_tolower.internal:
 	ld	a, l
 	sub	a, 'A'
 	cp	a, 1+'Z'-'A'
 	ret	nc
 
-	add	a, 'a'
-	ld	l, a
```
mateoconlechuga, line 23: How does this work

mateoconlechuga (approved):

src/libc/tolower.src
calc84maniac: Lowercase and uppercase letters in ASCII differ only by bit 5. Incidentally, that's also how the optimized loops for str[n]casecmp work by doing an xor check between the two characters.

PR #483, opened 2024-03-29
Allow override of $PREFIX

myclevorname:
I tried to run `make install`, but it kept installing to ./CEdev. I know the docs say it is supposed to install to ~/CEdev, but it didn't install there. I tried setting PREFIX, but the same result occured. This should fix that.


myclevorname:
From the IRC chat, it looks like some of the behavior is intentional for the CI.

PR #489, opened 2024-06-11
Added additional macros/constants to sys/lcd.h and created sys/spi.h

ZERICO2005:
I wanted to add some additional macros/constants to `sys/lcd.h`; such as being able to set RGB/BGR and Bits per pixel with `lcd_VideoMode = LCD_BGR565` using the lower 16bits of `lcd_Control`, or just the Bits per pixel using `lcd_VideoBPP = LCD_INDEXED8` to set the lower 8bits of `lcd_Control`.

Additionally, I thought it would be nice to add `#define lcd_Ram8 ((uint8_t*)0xD40000)` and `#define lcd_Ram16 ((uint16_t*)0xD40000)` to `sys/lcd.h`.

I also added `sys/spi.h`, which contains #define's for the SPI registers, along with some macro routines to send SPI commands and parameters (based of the SPI sending example from [WikiTI](https://wikiti.brandonw.net/index.php?title=84PCE:Ports:D000#Sending)). There are also some macros for Inverting the colors, and for setting Row-Major/Column-Major through the SPI commands.

### DIFF
--- a/src/ce/include/sys/spi.h
+++ b/src/ce/include/sys/spi.h
@@ -1,0 +1,104 @@
+/**
+ * @file
+ * @authors
+ * @brief FTSSP010 SPI controller define file
+ */
+
+#ifndef SYS_SPI_H
+#define SYS_SPI_H
+
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* @cond */
+#define spi_ControlRegister0     ((volatile void*)0xF80000)
+#define spi_ControlRegister1     ((volatile void*)0xF80004)
+#define spi_ControlRegister2     ((volatile void*)0xF80008)
+#define spi_StatusBits           ((const volatile void*)0xF8000C)
+#define spi_InterruptControl     ((volatile void*)0xF80010)
+#define spi_InterruptStatus      ((const volatile void*)0xF80014)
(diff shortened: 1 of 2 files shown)


src/ce/include/sys/spi.h
```diff
@@ -0,0 +1,104 @@
...
+    *(volatile uint8_t*)spi_FIFO = ((x) >> 3) & 0b111; \
+    *(volatile uint8_t*)spi_FIFO = (x) & 0b111; \
+    *(volatile uint8_t*)spi_ControlRegister2 = 0x1; \
+    while ( ((const volatile uint8_t*)spi_StatusBits)[1] & 0xF0) {}; \
+    while ( ((const volatile uint8_t*)spi_StatusBits)[0] & (1 << 2)) {}; \
+    *(volatile uint8_t*)spi_ControlRegister2 = 0x0; \
+} while(0)
+
+/** @todo Implement vsync */
+#define SPI_UNINVERT_COLORS() SPI_COMMAND(0x20)
```
adriweb, line 69: Maybe `SPI_LCD_*` instead of `SPI_*` for those?

ZERICO2005: Now that calc84maniac has worked on `lcddrvce`, it might be more appropriate to remove some of the SPI code

mateoconlechuga:
Do we still need this

PR #505, opened 2024-10-02
Fixed ldexpf behaviour for NaN and Inf (e.g. ldexpf(Inf, -1))

ZERICO2005:
ldexpf currently checks that `current_exponent + add_exponent <= exponent_max`. However, subtracting from the exponent value of Inf or NaN will result in a finite value. I have fixed this so that Inf and NaN are returned unmodified.


mateoconlechuga:
Thanks!

PR #506, opened 2024-10-07
Add Nix flake

myclevorname:
Here are some differences in the flake people might need to know.

1. The comment in the resulting programs is "CE C Toolchain" by default and not the current time to ensure reproducibility.
2. fasmg, conv*, and ez80-* are not included in the bin/ folder, but I overrode the paths to said programs in `src/makefile.mk` to be absolute paths to the actual programs within the Nix store. If you want access to those, use a devshell, like the one in `examples/nix-flake`, which stores a template for creating a *new* program with Nix.
3. The `cedev-nightly` package has been tested, but the `cedev-stable` package hasn't been tested.
4. When writing CE C programs with Nix, you can manually toggle to the stable or nightly toolchains using the `stableToolchain` flag.
5. Instead of typing `nix build .#cedev-nightly`, you have to type `.?submodules=1#cedev-nightly` because Nix doesn't fetch submodules by default for some odd reason.
6. The Flake file looks like a mess and may need to be cleaned up.


.gitignore
```diff
@@ -61,3 +61,8 @@ src/include/*.alm
 src/include/ez80.inc
 src/include/ti84pceg.inc
 src/include/tiformat.inc
+
+# Nix files
+examples/nix_flake/flake.lock
```
Eeems, line 66: `nix_flake` or `nix-flake`?

myclevorname: Oh yeah

mateoconlechuga:
There are so many things wrong here.

PR #508, opened 2024-10-10
Add _fpclassifyf and _isnormalf

ZERICO2005:
Added the `_fpclassifyf` function and the `fpclassify` macro along with its macro constants. Also added `_isnormalf`

```c
#define FP_NORMAL    0x1
#define FP_ZERO      0x2
#define FP_SUBNORMAL (FP_NORMAL | FP_ZERO)     /* 0x3 */
#define FP_INFINITE  0x4
#define FP_NAN       (FP_NORMAL | FP_INFINITE) /* 0x5 */
```


mateoconlechuga:
Thanks! I think we'd want to get these into assembly form

PR #509, opened 2024-10-10
Improved the precision (and probably speed) of cbrtf.

ZERICO2005:
Minimum precision: 21.415037 bits at both +4.251052856e+02 and +7.969426579e+17 for cbrtf.

Precision is calculated as `log2(fabs( true_cbrt(x) - approx_cbrt(x) )) - ilogb(fabs( true_cbrt(x) ))`

The previous method of `powf(x, 1.0f / 3.0f)` had a minimum precision of 21.0 bits at +4.037431946e+02, and a minimum precision 19.299560 bits at +1.187444200e+07, slowly losing precision at larger magnitudes.

Precision tested with 32bit floats on x86_64

### DIFF
--- a/src/libc/cbrt.c
+++ b/src/libc/cbrt.c
@@ -1,18 +1,30 @@
+#include <stdint.h>
 #include <math.h>
 
-/**
- * @remarks Newtons iteration will typically divide by 3.0, but I found that
- * multiplying by ~0.3328 gives a more accurate result somehow. This is rather
- * convienient since multiplication is faster than division.
- */
-#define magic_third 0.3328f
+typedef union Float32_Bitwise {
+	float flt_part;
+	uint32_t u32_part;
+} Float32_Bitwise;
 
 /**
- * @remarks Minimum precision: 21.415037 bits at both +4.251052856e+02 and
- * +7.969426579e+17 for cbrtf.
- * The previous method of powf(x, 1.0f / 3.0f) had a minimum precision of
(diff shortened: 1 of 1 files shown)


src/libc/cbrt.c
```diff
@@ -1,8 +1,53 @@
...
+ */
+float cbrtf(const float x)
 {
-    return powf(x, 1/3.f);
+	if (x == 0.0f || !isfinite(x)) {
+		// Perserves signed zeros, inf, and NaN
+		return x;
+	}
+	const float fabs_x = fabsf(x);
+	const float fabs_x_mul_2 = 2.0f * x;
```
calc84maniac, line 25: Shouldn't this be using fabs_x instead of x?

mateoconlechuga:
`powf(x, 1/3.f);` is smaller tho

calc84maniac:
> `powf(x, 1/3.f);` is smaller tho

It's also semantically different (especially for negative inputs), so if someone wants to save space (assuming they're already using powf elsewhere) they can call powf like you just specified.

ZERICO2005:
> `powf(x, 1/3.f);` is smaller tho

True, although it should probably be changed to this
```c
if ( x == 0.0f || !isfinite(x) ) {
	return x;
}
return copysignf(powf(fabsf(x), 1/3.f), x);
```

ZERICO2005:
I will close this pull request as calc84maniac has written an assembly version of this function.

PR #513, opened 2024-11-07
Optimize graphx clipping logic

calc84maniac:
Reworks sprite clipping to use range-based (unsigned) comparisons instead of signed comparisons, and also moves shared pointer calculation logic into the clipping routine. Additionally, logic is added to gfx_SetClipRegion to apply SMC to all clipping routines. This saves ~~178~~ 215 bytes overall, and I measured the following changes in cycle counts for gfx_Sprite (up until the gfx_Wait for drawn sprites, or through the return for culled sprites):

fully onscreen: 1177 -> ~~716~~ 644

clipped top: 1450 -> ~~860~~ 772
clipped left: 1488 -> ~~864~~ 780
clipped right: 1254 -> ~~756~~ 684
clipped bottom: 1202 -> ~~744~~ 672

clipped top left: 1761 -> ~~1008~~ 908
clipped top right: 1527 -> ~~900~~ 812
clipped bottom left: 1513 -> ~~892~~ 808
clipped bottom right: 1279 -> ~~784~~ 712

clipped left+right: 1565 -> ~~863~~ 779
clipped top+bottom: 1475 -> ~~859~~ 767
clipped top+left+right+bottom: 1863 -> ~~1006~~ 902

culled top: 471 -> ~~370~~ 334
culled left: 803 -> ~~540~~ 468
culled right: 776 -> ~~540~~ 468
culled bottom: 474 -> ~~370~~ 334

I'll probably take a look at gfx_RLETSprite later and add it to this PR if I can improve it.

### DIFF
--- a/src/graphx/graphx.asm
+++ b/src/graphx/graphx.asm
@@ -239,7 +239,47 @@ macro setSmcBytesFast name*
 	jp	(hl)
 end macro
 
+macro setSmcBytesInline name*
+	local temp, list
+	postpone
+		temp equ each
+		irpv each, name
+			temp equ temp, each
+		end irpv
+		list equ temp
+	end postpone
+
+	match expand, list
+		iterate expand
+			ld	(each),a
+		end iterate
+	end match
+end macro
+
+macro setSmcWordsInline name*
+	local temp, list
+	postpone
+		temp equ each
+		irpv each, name
+			temp equ temp, each
+		end irpv
+		list equ temp
+	end postpone
+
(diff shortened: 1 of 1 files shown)


calc84maniac:
While updating all clipped routines to use SMC, I found and fixed two bugs in flood fill clipping, the first being that X clipping on the left was always treated as 0, and the second being that X clipping on the right was off by one (treating the upper bound as inclusive instead of exclusive).

PR #517, opened 2024-11-08
Fixed the precision of expm1(float x) when `x` is small.

ZERICO2005:
The accuracy of expm1f(float x) has been fixed. It will `return x` when `x` is small for 12.0bits of precision, or it can `return x + 0.5f * (x * x)` when `x` is small for 16.0bits of precision. It falls-back to the naive `return expf(x) - 1.0f` when `x` is large.


src/libc/expm1.c
```diff
@@ -1,8 +1,24 @@
 #include <math.h>
 
-float expm1f(float x)
+float expm1f(const float x)
```
adriweb, line 3: Why? That's not needed or in the standard.

ZERICO2005: Adding const to parameters passed by value doesn't change the ABI. It was a micro optimization I added, although it probably doesn't change the compiler output

mateoconlechuga:
Can you please clean up this commit (remove the `const` and the `#if 1`) code? Thanks

PR #520, opened 2024-11-18
Update math_test example with 48-bit math

runer112:


runer112:
Note: 48-bit not doesn't work properly (https://github.com/jacobly0/llvm-project/issues/48), but the auto-test hash is set according to this incorrect value so auto-testing doesn't fail.

PR #523, opened 2025-01-05
Optimize is*

myclevorname:
Note: I haven't tested these optimizations


mateoconlechuga:
Then why should I even bother looking at this

adriweb:
(The problem being that we've been bitten several times in the past by untested things.... including again today, because we're too nice trusting everything 😅)
(The CI could be improved for external PRs, I think it's only for branches here that it gets tested... for rom related reasons, it's kind of annoying)

PR #527, opened 2025-01-10
Integrated 64bit long double into the toolchain

ZERICO2005:
Integrated a 64bit `long double` type into the toolchain, implemented using [Berkeley SoftFloat 3e](http://www.jhauser.us/arithmetic/SoftFloat.html). SoftFloat is under `toolchain/src/softfloat` and `CEdev/lib/softfloat` currently. All CRT `long double` ABI calls have been implemented, along with some `math.h` functions.

The following routines are implemented with SoftFloat:
```
_dadd _dsub _dmul _ddiv _drem
_ftod _ltod _lltod _ultod _ulltod
_dtof _dtol _dtoll _dtoul _dtoull
fmal sqrtl fmodl modfl
truncl floorl ceill roundevenl nearbyintl
roundl lroundl llroundl
rintl lrintl llrint
```

I have also added the C23 `int iszero(real-floating x)` function to `math.h`. I also intend on adding `issubnormal`, `nextup`, and `nextdown` from C23.


ZERICO2005:
long double float64 appears to work in both CEmu and on the calculator.
Summary of changes since January 11th:
* All the float64 and float32 classification routines have been tested.
* float64 arithmetic and conversion works.
* optimized some routines in assembly.
* libcalls will directly call softfloat functions instead of going through a C wrapper function.
* ilogbl, frexpl, and ldexpl round correctly and properly handle denormal values.
* nextafter, nextup, nextdown, nexttowards are implemented for both float32 and float64
* strtold was added (based off the current strtof routine)
* added `-Wimplicit-float-conversion -Wimplicit-int-float-conversion` to catch bugs in libc functions
* scalbn now aliases ldexp since FLT_RADIX is 2
* Fixed fmaxf, fminf, and fdimf to handle NaNs/signed-zeros correctly

Mathematical functions (sinl, expl, tgammal, etc) and remainderl/remquol will be implemented later on.

ZERICO2005:
Everything should be good to merge. The one potential problem was outlined here https://github.com/CE-Programming/toolchain/issues/540. So I've made a temporary fix to allow for the `long double` routines to link correctly
```c
// <math.h>
long double _debug_fabsl(long double x);
#define fabsl _debug_fabsl
long double _debug_fmaxl(long double x, long double y);
#define fmaxl _debug_fmaxl
```
```asm
; fabsl.src
	public	_fabsl, __debug_fabsl
__debug_fabsl:
_fabsl:
	; ...
```
```c
// fmax.c
#include <math.h>
#ifdef fmaxl
#undef fmaxl
#endif
long double fmaxl(long double x, long double y) {
    // ...
}
long double _debug_fmaxl(long double, long double) __attribute__((alias("fmaxl")));
``

PR #530, opened 2025-01-15
Added SPI register defines

ZERICO2005:
Defines for the SPI registers. Not sure if this should go inside `lcddrvce.h` or in its own header.


ZERICO2005:
I'll commit the SPI code to https://github.com/CE-Programming/toolchain/pull/489

PR #531, opened 2025-01-17
Use assembly version of feof

myclevorname:
The code was generated by switching the optimization flag to -Ofast, which automatically uses IY. After a few attempts, I couldn't write a faster function, so I just included that.

### DIFF
--- a/src/libc/feof.c
+++ b/src/libc/feof.c
@@ -1,7 +1,0 @@
-#include <stdio.h>
-#include <fileioc.h>
-
-int __attribute__((weak)) feof(FILE *stream)
-{
-    return stream->eof;
-}
--- a/src/libc/feof.src
+++ b/src/libc/feof.src
@@ -1,0 +1,11 @@
+	assume	adl = 1
+
+	section	.text
+	weak	_feof
+_feof:
+	ld	iy, 0
+	add	iy, sp
+	ld	iy, (iy + 3)
+	sbc	hl, hl
+	ld	l, (iy + 1)
+	ret


mateoconlechuga:
No, this is not maintainable

debrouxl:
This kind of optimization might be defensible (if not necessarily desirable) from a maintenance standpoint if that `+ 1` in the last `ld` could somehow be written as the equivalent of `offsetof(FILE, eof)`, against an assembler which understands C structs in header files it is instructed to #include. That way, if the FILE struct ever changed, the assembly code would remain unchanged, and the generated binary would be correct.

However... if the efficiency of `feof()` isn't good enough size-wise and/or speed-wise, which I from the outside suspect is the motivation for this work... isn't the real problem here that a trivial getter function is out of line ?
I went to check in GCC4TI's headers ( https://github.com/debrouxl/gcc4ti/blob/next/trunk/tigcc/include/C/stdio.h ), and `feof()` is a macro. On the 68000, which is the platform I've written the most assembly code for, macro / static inline function is the correct thing to do for avoiding an out of line call and a test; likewise on x86, which I mainly read, and ARM, which I haven't written for years.
I'm aware that the Z80 and eZ80 are more register-starved than all of these ISAs, but whichever the ISA, pushing argument, calling a function, and popping argument isn't free, even notwithstanding the fact that a call to an out of line function breaks optimized code generation in other ways.

mateoconlechuga:
It's a weak function @debrouxl

adriweb:
Apparently it's common to have it implemented as a macro, from what I can see here and there, and indeed it would make more sense if the compiler doesn't already inline it in real code.

it would be a better optimization overall than a very small out-of-line ASM change, anyway, I suppose.

ZERICO2005:
```diff
int __attribute__((weak)) feof(FILE *stream)
{
+    if (stream == NULL || stream == stdin || stream == stdout || stream == stderr)
+    {
+        return 0;
+    }
    return stream->eof;
}
```
And `feof` already had bugs to begin with

PR #533, opened 2025-01-28
Implemented i48abs(int48_t) and added prototypes for i48div

ZERICO2005:
`signed __int48 i48abs(signed __int48 n)` is implemented in assembly. The function prototypes for `i48div_t i48div(signed __int48 numer, signed __int48 denom)` have been declared, but the function is not implemented.


mateoconlechuga:
Where did `__int48` come from?

calc84maniac:
> Where did `__int48` come from?

It's defined by the compiler. I recommended using the compiler definition rather than including stdint.h to avoid adding unnecessary include file dependencies (which can cause users to accidentally omit include files they should be including).

PR #537, opened 2025-02-03
Added __frem to CRT (wraps fmodf)

ZERICO2005:
```c
float fwrap(float x, float min, float max) {
    if (min > max)
        return fwrap(x, max, min);
    return (x >= 0 ? min : max) + fmodf(x, max - min);
}
```
This code when compiled with `-Ofast` (in C) would emit a lib-call to `__frem` instead of a call `_fmodf`. However, `__frem` wasn't defined in the toolchain. This has been fixed by adding the CRT routine for `__frem`, which wraps `_fmodf` for now.

Credit to @merthsoft for finding this bug

### DIFF
--- a/src/crt/frem.src
+++ b/src/crt/frem.src
@@ -1,0 +1,21 @@
+	assume	adl=1
+
+	section	.text
+
+	public	__frem
+
+; wraps _fmodf
+__frem:
+	push	de, hl
+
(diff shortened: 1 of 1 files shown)


adriweb:
(retval fixed by 82f472c)

PR #546, opened 2025-03-02
added perror and strerror along with autotests for them

ZERICO2005:
also adds `ti_routines.src`, which allows for the explicit use of ti's `sprintf` to reduce binary size


ZERICO2005:
using pr/str_error instead

PR #547, opened 2025-03-03
Seperated math.h functions into funcf.* and funcl.*, and corrected the formatting of math.h functions

ZERICO2005:
It doesn't make sense to have `erfc.c` contain both `erfcf(float)` and `erfcl(long double)`, especially when different algorithms may be used for the different precisions. So I split the `<math.h>` functions such that `funcf.*` contains `float`, `funcl.*` contains `long double`, and `func.*` contains both `float` and `long double`.

I also fixed `softfloat` in `linker.mk` as Adriweb pointed out to me. Additionally, the formatting of the `<math.h>` functions and the Zilog copyright headers were corrected.

Otherwise, I patched a softlocking bug in `lgammaf` (`lgammaf(-1.0e8f)` would perform 10^8 iterations and etc), corrected the behavior for negative inputs in `erfc`, along with other minor bug fixes.


ZERICO2005:
I fixed two typos inside of comments
```diff
-/*                Copyright (C) 2000-2008 by Zilog, Inc.                */
+/*                 Copyright (C) 2000-2008 Zilog, Inc.                  */
-* strtold - string to float conversion
+* strtold - string to long double conversion
```

PR #550, opened 2025-03-07
libc: add VRAM2 allocator.

adriweb:
Used in prod in KhiCAS nowadays, thought it'd be better to upstream it.

### DIFF
--- a/src/libc/allocator.src
+++ b/src/libc/allocator.src
@@ -33,13 +33,17 @@ _malloc := __simple_malloc
 _free := __simple_free
 _realloc := __simple_realloc
 
-end if
-
-if defined ALLOCATOR_STANDARD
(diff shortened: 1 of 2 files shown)


mateoconlechuga:
lol what, none of this makes sense

adriweb:
lol what do you mean?
It definitely works, at least, so it must make some sense :P

That said I only cleaned up the formatting, not the algorithm it uses.

adriweb:
After more internal brainstorming, this can actually stay inside KhiCAS while still not require a modification of the toolchain (see ALLOCATOR_CUSTOM), which was the original goal anyway.

Regardless, having a toolchain-provided way to "add memory areas" to the heap is a good feature so it may get added at some point.

PR #556, opened 2025-03-15
Add API for LCD hardware cursor

nitinseshadri:
This PR adds API to make interacting with the LCD's hardware cursor easier.


mateoconlechuga:
You need to wrap most of these functions in a `do {} while (0)` construct.

nitinseshadri:
done!

adriweb:
Did you have any demo/example code that uses this btw? :)

nitinseshadri:
[PETSCII Robots CE](https://github.com/nitinseshadri/ce-robots) uses this to display search, move, and use cursors. The only difference is that it is implemented as functions in a header file (`cursor.h`) as opposed to macros here in this PR. You can have PETSCII Robots use the macros through a find and replace of instances of `cursor_` to `lcd_Crsr` in PlatformCE.cpp (e.g. `cursor_Setup()` would become `lcd_CrsrSetup()`).

mateoconlechuga:
I think the question was if you could add an example to the toolchain so other people can have a reference on how to use it 😉

nitinseshadri:
Got it, I added an example to the PR.

calc84maniac:
I think it would also be good to have methods to set the cursor clip amount and the cursor image index, as well as a way to explicitly set the data for a 32x32 or 64x64 image (the former allowing to specify an image index). You could still keep the existing generic data set method that checks the current cursor size and implies image index 0, if that makes things easier. Also accordingly, the cursor show/hide methods should be updated to avoid overwriting the image index (for example, with `lcd_CrsrCtrl |= 1;`)

nitinseshadri:
done!

Issue #543, opened 2025-02-22
Add asprintf vasprintf


ZERICO2005:
It would be nice to add `asprintf` and `vasprintf`. They are similar to `sprintf`, but they will allocate the buffers for you
```c
int asprintf(char **__restrict strp, const char *__restrict fmt, ...);
int vasprintf(char **__restrict strp, const char *__restrict fmt, va_list ap);
```
https://www.man7.org/linux/man-pages/man3/asprintf.3.html

mateoconlechuga:
No

adriweb:
Elaborate?

After all, they are in some TS extension Dynamic memory extensions
<http://en.cppreference.com/w/c/experimental/dynamic> of the C standard,
and other libc impls have them

Could be useful especially when porting persons that use them and it would
only get linked when used so 🤷‍♂️

ZERICO2005:
Without `asprintf`, I have to do the alternative. Getting the length of the string with `snprintf(NULL, 0, ...)`, `malloc`'ing `length + 1` bytes, and then writing to the buffer. There is a lot of boilerplate involved for handling errors, it is easy to forget to allocate `length + 1` bytes, and etc. which is why I think it is best implemented as a library function.
```c
int len_str = snprintf(NULL, 0, /* args */);
if (len_str < 0) {
	return NULL;
}
size_t str_size = (size_t)len_str + 1;
char* buf = (char*)malloc(str_size);
if (buf == NULL) {
	return NULL;
}
int status = snprintf(buf, str_size, /* args */);
if (status < 0) {
	free(buf);
	return NULL;
}
return buf;
```

PR #558, opened 2025-03-20
Added asprintf and fprintf along with ti_sprintf macros

ZERICO2005:
Functions added:

* `fprintf`/`vfprintf`
* `asprintf`/`vasprintf`
* `ti_sprintf`,`ti_snprintf`, and `ti_asprintf` to `<ce/include/ti_sprintf.h>`

`fprintf` redirects the output of `nano_printf` to `fputc`

`asprintf` is implemented via two calls to `snprintf` https://github.com/CE-Programming/toolchain/issues/543

`ti_sprintf` makes explicit calls to the C89 `sprintf` from OS_LIBC. This can reduce binary sizes significantly compared to `nano_printf`

`ti_snprintf` and `ti_asprintf` write to unmapped memory `0xFC1000` to get the size of the output. Writing to `0xFB0000` also works, but will output to the CEmu console. Otherwise, writing to `0xFC1000` should be okay unless the output exceeds ~258000 characters. Because `ti_vsprintf` doesn't exist, they are both implemented as macros, and will evaluate `__VA_ARGS__` twice.

`ti_asprintf` will `malloc` a buffer that `ti_sprintf` will write to, returning `NULL` on `malloc` failure. The buffer shall be deallocated with `free`

`ti_snprintf` will return an empty string if the result of `ti_sprintf` won't fit inside the buffer. This was done to make the routine more compact, as implementing the truncation behavior of `snprintf` would require calls to `malloc`. If the user wants the output to be truncated, they can use `ti_asprintf` and `strncpy`
```c
char buf[20];
char* temp;
ti_asprintf(&temp, format, ...);
if (temp != NULL) {
	strncpy(buf, temp, sizeof(buf));
	free(temp);
}
```


ZERICO2005:
Tests for fprintf, asprintf, ti_sprintf, ti_snprintf, and ti_asprintf have been added now.
Also added `stpcpy` and `memccpy` from @calc84maniac along with tests for them.

mateoconlechuga:
I am a bit worried that the `ti_` prefix may conflict with the naming convention fileioc.

adriweb:
Gotta trust the tests 🫡

ZERICO2005:
> I am a bit worried that the `ti_` prefix may conflict with the naming convention fileioc.

The functions I've added use this naming convention `ti_function`, while the ones in fileioc use `ti_Function`. I guess they could also be `ce_sprintf` or etc

ZERICO2005:
the function naming should be okay as is, and won't conflict

mateoconlechuga:
lol wtf

adriweb:
I'm not sure this is an issue now but if we want to change the naming convention to ce_xxx instead, we can always do that before a release anyway, no worries. Do you think it's better than ti_xxx ?

PR #560, opened 2025-03-21
float classification is now constexpr, added sinpi/cospi functions

ZERICO2005:
changes:
* float classification can now be done at compile time in C++. This only worked when I added `constexpr` to the function prototype in C++, so I'm not sure if it can also be done for C.
* Added `__NOEXCEPT __attribute__((const))` to non-standard function prototypes, such as `_fpclassifyf`
* optimzied the float classification routines. Most of them return bool (which gets stored in the carry flag from what I've seen) instead of int
* added sin(pi * x), cos(pi * x), and tan(pi * x) from C23, which allow for accurate range reduction. I didn't add `asinpi acospi atanpi atan2pi` since they are trivial to implement `atan2(y, x) / pi`

### DIFF
--- a/src/libc/include/__math_def.h
+++ b/src/libc/include/__math_def.h
@@ -110,6 +110,10 @@ double      cos(double);
 float       cosf(float);
 long double cosl(long double);
 
+double      cospi(double);
+float       cospif(float);
+long double cospil(long double);
+
 double      cosh(double);
 float       coshf(float);
 long double coshl(long double);
@@ -289,6 +293,10 @@ double      sin(double);
 float       sinf(float);
 long double sinl(long double);
 
+double      sinpi(double);
+float       sinpif(float);
+long double sinpil(long double);
+
 double      sinh(double);
 float       sinhf(float);
 long double sinhl(long double);
@@ -300,6 +308,10 @@ long double sqrtl(long double);
 double      tan(double);
 float       tanf(float);
 long double tanl(long double);
+
+double      tanpi(double);
+float       tanpif(float);
+long double tanpil(long double);
 
 double      tanh(double);
 float       tanhf(float);
--- a/src/libc/include/__math_def.h
+++ b/src/libc/include/__math_def.h
@@ -1,6 +1,7 @@
 #ifndef _MATH_DEF_H
 #define _MATH_DEF_H
 
+#include <cdefs.h>
 #include <stdbool.h>
 
 #define NAN          __builtin_nanf("")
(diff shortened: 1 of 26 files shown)


runer112 (commented):
> float classification can now be done at compile time in C++. This only worked when I added `constexpr` to the function prototype in C++, so I'm not sure if it can also be done for C.

Could you create tests for this using `static_assert`?

src/libcxx/include/cmath
```diff
@@ -12,61 +12,198 @@ namespace std {
 using ::float_t;
 using ::double_t;
 
+/* Clang 20.1.0 required for __builtin_signbit to work in constexpr */
+#if 0
+
+inline constexpr bool signbit(float __x) {
+    if (__builtin_constant_p(__x)) {
+        return __builtin_signbitf(__x);
+    }
```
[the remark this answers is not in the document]
ZERICO2005: that seems to work. `(__x < 0.0)` works in most cases except when `__x` is signed zero or NaN. I came up with this code to test for signed zero, but I am not sure how to handle signed NaN
```c++
inline constexpr bool signbit(double __x) {
    if (__builtin_constant_p(__x)) {
        if (__x == 0.0) {
            return (1.0 / (__x) < 0.0); // turns +-zero into +-infinity
        }
        return (__x < 0.0);
    }
    return static_cast<bool>(_signbitf(__x));
}
```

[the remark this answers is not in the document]
ZERICO2005: Got it fixed. `__builtin_copysign` works at compile time
```c++
inline constexpr bool signbit(double __x) {
    if (__builtin_constant_p(__x)) {
        return (__builtin_copysign(1.0, __x)  < 0.0);
    }
    return static_cast<bool>(_signbitf(__x));
}
```

src/libc/include/__math_def.h
```diff
@@ -40,30 +41,30 @@ typedef double double_t;
...
-int _isnormall(long double n);
-int _isfinitel(long double n);
-int _iszerol(long double n);
-int _issubnormall(long double n);
+int _isinfl(long double) __NOEXCEPT_CONST;
+int _isnanl(long double) __NOEXCEPT_CONST;
+bool _isnormall(long double) __NOEXCEPT_CONST;
+bool _isfinitel(long double) __NOEXCEPT_CONST;
+int _iszerol(long double) __NOEXCEPT_CONST;
+int _issubnormall(long double) __NOEXCEPT_CONST;
```
runer112, line 59: While we're here, could we optimize the remaining fp classification functions to align with the majority and return `bool`?

ZERICO2005: I can take a look at it. They're a bit trickier since they return more than once

runer112 (changes requested):

src/libc/isnanl.src
runer112, line 1: I can't make this suggestion inline, but please update the function signature in the comment:

```
; bool _isnanl(long double)
```

runer112 (changes requested):

mateoconlechuga:
Looks good to me for a merge, anyone else?

runer112 (approved):
Provided

PR #561, opened 2025-03-23
Implement strlcpy

nitinseshadri:
This PR adds an implementation and tests for `strlcpy`, and corrects its signature in string.h. `strlcpy` is a BSD variant of strncpy that ensures the destination is null-terminated. More info about strlcpy here: https://linux.die.net/man/3/strlcpy

### DIFF
--- a/src/libc/include/string.h
+++ b/src/libc/include/string.h
@@ -44,7 +44,7 @@ char *stpcpy(char *__restrict dest, const char *__restrict src)
 char *stpncpy(char *__restrict dest, const char *__restrict src, size_t n)
     __NOEXCEPT __attribute__((nonnull(1, 2)));
 
-char *strlcpy(char *__restrict dest, const char *__restrict src, size_t n)
+size_t strlcpy(char *__restrict dest, const char *__restrict src, size_t n)
     __NOEXCEPT __attribute__((nonnull(1, 2)));
 
 char *strcat(char *__restrict dest, const char *__restrict src)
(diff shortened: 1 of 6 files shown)


ZERICO2005:
One good thing to check for is that a call to `strlcpy` is emitted in `lto.src`. Sometimes Clang will replace it with its own implementation.

nitinseshadri:
Yep, I checked `lto.src` and there are calls to `strlcpy` where I expected them. I also specified `-O0` in the Makefile.

calc84maniac:
I was going to do a review, but as was mentioned in chat it would be good to unconditionally do the strlen at the start and avoid doing it twice. My other feedback for now is that `sbc hl,hl` is always more optimized than `ld hl,0`, even if you have to explicitly clear the carry flag first.

calc84maniac:
Did my own optimization pass which saves 14 bytes if I counted right, let me know if anything doesn't make sense or I somehow broke something. Felt that would be easier than trying to point out individual changes in a review

nitinseshadri:
Thanks, it makes sense to me, and the test program works properly too.

mateoconlechuga:
I'm not sure how I feel adding BSD-specific implementations. I can understand some POSIX functions, but an argument needs to be made before we decide to start pulling in random BSD stuff which clutters things and makes it hard for users to know what is actually available. I would prefer to only see C-standard compliant libc.

nitinseshadri:
`strlcpy` and `strlcat` were added to the latest POSIX standard, POSIX.1-2024: https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/string.h.html (referred to as "Issue 8" on this page)

ZERICO2005:
quite a few of these POSIX functions are already available as Clang builtins too https://github.com/llvm/llvm-project/blob/main/clang/include/clang/Basic/Builtins.td

PR #562, opened 2025-03-23
boot_sprintf docs, fixed fprintf

ZERICO2005:
Currently writing documentation for `boot_sprintf`

One thing to decide on is if `boot_snprintf` and `boot_asprintf` should be defined to `snprintf` and `asprintf` if `HAS_PRINTF = NO`

### DIFF
--- a/docs/headers/boot_sprintf.rst
+++ b/docs/headers/boot_sprintf.rst
@@ -1,0 +1,76 @@
+.. _boot_sprintf:
+
+boot_sprintf
+======================
+
+.. code-block:: c
+
+    #include <boot_sprintf.h>
+
+The OS comes with an implementation of ANSI C89 `sprintf`, which can reduce the size of a program by ~8 KiB. However, it is very limited in functionality. It does not support width specifiers, and it won't accept :code:`long` or :code:`float` arguments.
+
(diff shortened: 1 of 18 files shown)


adriweb:
Tested @ZERICO2005's [test program](https://pastebin.com/raw/HcwjA3x0) (with `boot_` instead of `ce_`) on 6 different HW revs, various 83/84/82, all OK.

adriweb:
KhiCAS also now links correctly.

ZERICO2005:
Summary: adds `boot_sprintf boot_snprintf boot_asprintf` and some documentation, fixes fprintf linking issues, and fixed the header bug in `<climits>` that was spotted by ssyntact

adriweb (changes requested):
OK so after changing from `0xFB0000` to `0xE40000` in the latest push, which we thought would have been even better, I noticed it actually started to crash on M+ devices (with the test code still having `boot_sprintf`'s destination as `0xFB0000`), so maybe we'll have to go back (or use yet another address).

Edit: this was when launched from arTIfiCE. If launched through AsmHook, I can't get it to crash at all.

This brings up two interesting points though:
- CEmu didn't handle this (did not crash), @calc84maniac is looking into it.
- We need to run tests on both pre-M and M+ ROMs! I can take care of this.

mateoconlechuga:
This PR was *NOT* ready to be merged.

PR #563, opened 2025-03-25
changed boot_sprintf.h to ti/sprintf.h and fixed the documentation

ZERICO2005:
Fixed the documentation for `boot_sprintf`

### DIFF
--- a/docs/headers/ti/index.rst
+++ b/docs/headers/ti/index.rst
@@ -23,6 +23,7 @@ whoever in the community first discovered something gave it a name following the
     info
     python
     real
+    sprintf
     tokens
     ui
     vars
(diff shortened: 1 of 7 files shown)


adriweb:
I fixed the syntax typo, but this remains to be fixed:

> **calc84maniac**:
>  so the doc says at the top it doesn't support width specifiers, but on the next line it says the width field is supported
>  did you mean length modifiers?
>
> **zerico2005**:
>  probably
>  `sprintf(%7d, val)` and `sprintf(%*d, width, val)` is what I meant by the width field

ZERICO2005:
I have fixed the formatting errors. Anything else that should be added to the documentation?

Issue #466, opened 2023-12-26
`__ultof` actually implements `__ltof`


runer112:
And `__ltof` implements a worse `__ltof`.

Example code demonstrating issue:

```c
#include <ti/getcsc.h>
#include <ti/screen.h>

#include <limits.h>
#include <stdio.h>

void testl(long x) {
  printf("%11li -> %11.0f", x, (float)x);
}

void testul(unsigned long x) {
  printf("%11lu -> %11.0f", x, (float)x);
}

int main() {
  os_ClrHome();

  testl(LONG_MAX);
  testl(LONG_MIN);
  testl(ULONG_MAX);
  os_NewLine();
  testul(LONG_MAX);
  testul(LONG_MIN);
  testul(ULONG_MAX);

  while (!os_GetCSC())
    ;
}
```

![image](https://github.com/CE-Programming/toolchain/assets/6608979/717b70c7-573e-401f-a81a-4ca575f65a47)

runer112:
`__ultof` and `__ltof` are both implemented as if `__fppack` accepts an unsigned 32-bit integer input, but it actually expects a signed 32-bit integer input.

runer112:
Discovered by 418cat: https://discord.com/channels/432891584451706892/432891584875593730/1188975936603947069.

ZERICO2005:
it also seems that `__fppack` does rounds up ties instead of rounding to even
`937097632ul` should round to even `937097600.0f`, not upwards to `937097664.0f`

PR #564, opened 2025-03-25
fixed ultof and added tests

ZERICO2005:
Fixes this issue https://github.com/CE-Programming/toolchain/issues/466

`ultof` now correctly handle values greater than `INT32_MAX`. However, ti's `ltof` will round to nearest with ties away from zero, while `ultof` will round to nearest with ties to even.

I am not sure if would be best for `ultof` to have the same rounding as `ltof`

Thanks to calc84maniac for help on `ultof.src`


mateoconlechuga:
ultof should have same rounding as ltof

ZERICO2005:
I have fixed it so it uses round to nearest with ties away from zero

PR #567, opened 2025-03-27
Optimized nanoprintf, and fixed boot_sprintf macro aliases

ZERICO2005:
implemented `npf_bufputc`, `npf_bufputc_nop`, `npf_putc_std`, and `npf_fputc_std` in assembly. This results in both smaller and faster routines

I also realized that `HAS_PRINTF` is not defined in C, which meant that the `boot_sprintf` macro aliases were always defined when including `<stdio.h>`
```c++
#ifndef HAS_PRINTF
# define snprintf boot_snprintf
# define asprintf boot_asprintf
#endif /* HAS_PRINTF */
```
This has now been fixed so they will only be enabled when HAS_PRINTF is defined and zero in C
```c++
#if defined(HAS_PRINTF) && HAS_PRINTF == 0
# define snprintf boot_snprintf
# define asprintf boot_asprintf
#endif /* HAS_PRINTF */
```

### DIFF
--- a/src/libc/fprintf.src
+++ b/src/libc/fprintf.src
@@ -10,7 +10,26 @@ if HAS_PRINTF
 _fprintf := __fprintf_c
 _vfprintf := __vfprintf_c
 
+	public	_npf_fputc_std
+_npf_fputc_std:
+;	call	__frameset0
+;	ld	hl, (ix + 6)
+;	ld	de, (ix + 9)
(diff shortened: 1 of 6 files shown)


mateoconlechuga:
um I really don't want to optimize or touch nanoprintf at all.

adriweb:
Well regardless of npf opts, the HAS_PRINTF fixes need to be merged in.

PR #573, opened 2025-04-01
use tabs

ZERICO2005:
they're better


ZERICO2005:
A surprising amount of .asm/.src files were modified from the "enhancements", so it would be good to fix those

mateoconlechuga:
no, just asm files use tabs

zwassall-gentuity:
IMO using tabs for indentation is a relic of the past. Spaces provide better layout flexibility and no ambiguity, and basically all modern IDEs will seamlessly perform indentation operations in space-indented files as if they were using tabs (so you don't have to go one space at a time).

runer112:
\<opinion>IMO using tabs for indentation is a relic of the past. Spaces provide better layout flexibility and no ambiguity, and basically all modern IDEs will seamlessly perform indentation operations in space-indented files as if they were using tabs (so you don't have to go one space at a time).\</opinion>

Regardless of my stance on the matter in general, in this particular case, I would think that people on both sides could agree after inspecting the diff that the bulk conversion has introduced a number of unappealing indentation changes and misalignments.

myclevorname (commented):
LGTM

PR #580, opened 2025-04-07
Added <ez80_builtin.h> and missing builtins

ZERICO2005:
Added `<ez80_builtin.h>`, which provides missing builtins
```
__ez80_clzc
__ez80_clzi48
__ez80_ctzc
__ez80_ctzi48
__ez80_ffsc
__ez80_ffss
__ez80_ffsi48
__ez80_popcountc
__ez80_popcounts
__ez80_popcounti48
__ez80_bswap24
__ez80_bswap48
__ez80_bitreverse24
__ez80_bitreverse48
__ez80_rotateleft24
__ez80_rotateleft48
__ez80_rotateright24
__ez80_rotateright48
```
While `__builtin_clzll(u48_value)` will often compile down to `call __i48ctzl`, `__ez80_ctzi48` provides an explicit way to call `__i48ctzl`, which can help guarantee behavior of `__i48ctzl(0) == 48`

The `clz`, `popcount`, `bitreverse`, and `bswap48` functions wrap CRT functions. `ctz` and `ffs` functions use `popcount(x ^ (x - 1))`, `i48add_1` and `i48sub_1` were added to help with this. rotate24/48 takes the mod 24/48 of an unsigned 8 bit shift value. `bswap24` uses an implementation I came up with, but should probably become a CRT function.

I also realized that the `_Generic` macros in `<stdbit.h>` won't work on non-functions such as builtins, so I have removed them for now.


mateoconlechuga:
Why are you adding comments like `unoptimized` and others. Rejected.

ZERICO2005:
Because I wasn't sure how to optimize those functions further. Some of them would require writing a dedicated ctz or ffs function to be optimized. Otherwise, all functions have been tested and verified across several inputs

mateoconlechuga:
Then the PR is in draft stage.

PR #590, opened 2025-04-16
Fixed float to integer punning in nanoprintf.c

ZERICO2005:
> union-cast is UB, let compiler optimize byte-copy loop.

The compiler wasn't able to see through the byte-copy loop so it generated inefficient code.
I have fixed this by using a memcpy to perform the float to integer punning. This reduced the required stack frame from 140 to 108 bytes, which allows the code to be within the +-127 byte range of `IX`


ZERICO2005:
This is actually a compiler bug that only occurs at `-Oz` I think. `-Og -O1 -O2 -O3 -Ofast -Os` are able to optimize the byte copy loop correctly.

Otherwise I am looking at updating nanoprintf, the newest version adds support for `%a %e %g` and might be smaller?

PR #593, opened 2025-04-21
Add LCDDRVCE library

calc84maniac:
Now you can do wacky LCD commands and stuff

I'd like some feedback on the API and docs, if possible. I consider the library implementation itself and its exported API to be more-or-less set in stone at this point, but the convenience macros can be changed as much as needed before merging.

I'm considering removing the bitfield-based raw parameter structs now that I've added parameter constants for most of the relevant commands. Though, that does bring to mind that we're missing a good way to define gamma parameters in C. Of course, that can be added later if needed.


mateoconlechuga:
This looks good to me

adriweb:
I've tried the two test programs on a few calcs:
- 82AEP (rev pre-A): both work fine
- 83PCE (rev L): both work fine
- 84+CEPy (rev S): both work fine
- 83PCEEP (rev O): half-res works, gradient "freezes" during the grey part at the top.
- 83PCEEP (rev S): half-res works, gradient "freezes" during the grey part at the top.

Here's the rev O "freezing" (I can still blindly press [clear] then turn the calc off then back on, though)
CleanShot 2025-05-22 at 20 56 30@2x

adriweb:
Merging due to it being ready as discussed earlier today:

> **Adriweb  — _07:09_**
> If everything goes well, should I merge?
> or do you still have some things to add
>
> **calc84maniac — _07:20_**
> I don't think there's anything else to add

PR #598, opened 2025-05-05
Optimized strcmp and strncmp. Also added memrchr

ZERICO2005:
Optimized `strcmp` and `strncmp`. Also added `memrchr`

`strncmp` has a linear speedup compared to Ti's routine (less clock cycles per character compared), making it much faster

`strcmp` has a constant speedup compared to Ti's routine (same clock cycles per character compared, but takes less clock cycles in total), which may be worth it when multiple short string comparisons need to be performed, such as when parsing text/scripts.

### DIFF
--- a/src/libc/memrchr.src
+++ b/src/libc/memrchr.src
@@ -1,0 +1,24 @@
+	assume	adl=1
+
+	section	.text
+
+	public	_memrchr
+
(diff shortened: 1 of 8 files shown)


mateoconlechuga:
Please fix the tests :)

PR #600, opened 2025-05-07
optimized float to integer routines

ZERICO2005:
Due to undefined behavior for out of range results, `dtol` and `dtoul` can be merged to be the exact same routine. Same applies for `dtoll` and `dtoull`.

I also undid the splitting of `ftoll` and `ftoull`. As `ftoll` already correctly handles all non-UB cases for both `(long long)float` and `(unsigned long long)float` to begin with.


mateoconlechuga:
I do not like how you are replacing the soft float routines like this.

PR #601, opened 2025-05-09
Optimized __setflag

ZERICO2005:
The compiler will emit `call pe, __setflag` after signed comparisons. I have added our own faster implementation of `__setflag` to the CRT.

```asm
__setflag:
	ret	po
; __invertflag:
	push	af
	ex	(sp), hl
	ld	a, l
	xor	a, $80	; invert S flag
	ld	l, a
	ex	(sp), hl
	pop	af
	ret
```
Since the compiler only appears to emit `call pe, __setflag` instead of `call __setflag`, it might be better to have it call `__invertflag` so it can skip `ret po`


mateoconlechuga:
This seems sort of risky to me

PR #609, opened 2025-05-21
Added a column-major graphics demo

ZERICO2005:
Based off the graphx buffered cube example

Uses code from graphy so vsync is properly handled and etc

I got errors from the `org` keyword used in `gfy_FillScreen`, so I've implemented it in C with memset instead for now.

### DIFF
--- a/examples/library_examples/lcddrvce/column_major/.gitignore
+++ b/examples/library_examples/lcddrvce/column_major/.gitignore
@@ -1,0 +1,7 @@
+obj/
+bin/
+src/gfx/*.c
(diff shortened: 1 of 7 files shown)


ZERICO2005:
I was not sure how to implement vsync in C, so I reused graphx/graphy functions for it. The purpose for this demo is to show how there is no diagonal screen tearing in column major mode

calc84maniac:
The point is that you shouldn't reuse graphx functions, this is a library demo so just call gfx functions directly

PR #615, opened 2025-06-09
Optimize graphx

ZERICO2005:
Optimized the clipping function for `gfx_Sprite`, `gfx_TransparentSprite`.
`_TransparentPlot` was unrolled 4 extra times for opaque plotting. This reduces the time for plotting 8 opaque pixels from `54F + 16R + 8W + 18` to `51F + 16R + 8W + 17`. Which makes it roughly 4% faster?
 
sprite flip/rotate90/rotate180 routines were optimized

other routines were also optimized, mainly in ways that reduce size while maintaining or improving performance. Although most performance gains were constant (i.e. 4F faster per function call) instead of linear (i.e. 2F * line_count faster)


ZERICO2005:
this is merging from the wrong branch, I'll fix that

PR #616, opened 2025-06-12
Optimize graphx

ZERICO2005:
Non-transparent sprite rotation was further optimized by a good amount.

Optimized the clipping function for `gfx_Sprite`, `gfx_TransparentSprite`.

Sprite flip/rotate90/rotate180 routines were optimized.

Other routines were also optimized, mainly in ways that reduce size while maintaining or improving performance. Although most performance gains were constant (i.e. 4F faster per function call) instead of linear (i.e. 2F * line_count faster).

### DIFF
--- a/src/graphx/graphx.asm
+++ b/src/graphx/graphx.asm
@@ -1992,14 +1992,15 @@ gfx_Line:
 	call	_ComputeOutcode
 	ld	(iy-2),a
 CohenSutherlandLoop:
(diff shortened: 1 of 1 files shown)


ZERICO2005:
Since I keep finding additional ways to save a byte or two, I will leave this as a draft for a bit longer. Otherwise everything should be stable now.

ZERICO2005:
forgot to remove some whitespace!

PR #618, opened 2025-07-08
Implemented dtof in assembly

ZERICO2005:
Wrote `__dtof` in assembly. Currently it raises no signals and is hard-coded to round-to-even.
* `FE_OVERFLOW` and `ERANGE` could be raised if a finite input overflows to infinity, or if the result was rounded to infinity (such as `(float)(FLT_MAX + 0.5ulp)`).
* `FE_UNDERFLOW` and `ERANGE` could be raised if a non-zero input becomes zero.
* `FE_INEXACT` could be raised if the result was rounded.

```
Convert BC:UDE:UHL F64 to E:UHL F32
Rounding: round to nearest with ties to even
Behaviour:
  Underflow: Returns signed zero. No signals raised.
  Subnormal: No signals raised.
  Rounded to Infinity: No signals raised.
  Overflow: Returns signed infinity. No signals raised.
  Signaling NaN: Quiet bit preserved. No signals raised.
  Quiet NaN: Quiet bit preserved. No signals raised.
  NaN Payloads: Copies the most significant payload bits. The LSB of mantissa is set if payload bits were discarded/truncated out.
```


ZERICO2005:
I suppose raising signals and errno could be implemented later on if needed.

PR #620, opened 2025-07-18
ldexpf now correctly handles/rounds subnormal inputs

ZERICO2005:
ldexpf/scalbnf now correctly handles/rounds subnormal inputs/outputs.

Since Ti's floats don't support negative zero, ldexpf also has a `__ldexpf_avoid_negative_zero` option (set to 1 currently), which guarantees that negative zero is only returned if the input is negative zero.

### DIFF
--- a/src/libc/ldexpf.src
+++ b/src/libc/ldexpf.src
@@ -16,183 +16,31 @@ _scalbn := _ldexpf
 
 else
 
-if 1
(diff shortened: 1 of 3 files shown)


ZERICO2005:
I suppose this implementation of `ldexpf` is fine for now. Later on we could use `ez80sf` routines to speedup the handling of subnormal values.

PR #621, opened 2025-07-19
Add clipped sprite rotation

ZERICO2005:
Adds `gfx_RotatedScaledSprite` and `gfx_RotatedScaledTransparentSprite` along with a demo under `toolchain/examples/library_examples/sprites_rotate_clipped`:

image

These functions return the size of the sprite before clipping is performed.

The performance of the clipped routines is identical to the unclipped routines apart from a small constant overhead to calculate the clipped region.


ZERICO2005:
Found a small optimization for `gfx_SetDefaultPalette`, and added an autotest for it

adriweb:
Would be the last merge before toolchain v13 release. If someone has time for a review :)

PR #623, opened 2025-08-11
Added a column-major demo for lcddrvce

ZERICO2005:
Rewrite of the previous attempt: https://github.com/CE-Programming/toolchain/pull/609

This demo shows how `lcddrvce` can be used to set the screen to column-major mode. It is similar to the `graphx/buffered_cube` example, but shows how column-major mode can be used to remove diagonal screen tearing.

Not exactly sure what comments I should add to the line drawing code.

### DIFF
--- a/examples/library_examples/lcddrvce/column_major/src/main.c
+++ b/examples/library_examples/lcddrvce/column_major/src/main.c
@@ -1,0 +1,195 @@
+#include <stdint.h>
+#include <stdlib.h>
+#include <math.h>
+#include <ti/getcsc.h>
+#include <graphx.h>
+#include <lcddrvce.h>
+
+#define ORG (-50)
+#define DEG (6.0)
+#define ANG (DEG * M_PI / 180.0)
+
+typedef struct {
+    float sin_ang;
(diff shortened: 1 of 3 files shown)


examples/library_examples/lcddrvce/column_major/src/main.c
```diff
@@ -0,0 +1,198 @@
...
+    }
+    int dD = 2 * dY - dX;
+    const int dD_jump = 2 * (dY - dX);
+    dY *= 2;
+
+    uint8_t* buffer = (uint8_t*)lcd_LpBase + (x0 * LCD_HEIGHT) + y0;
+    const uint8_t color = line_color;
+
+    uint24_t x = (uint24_t)(x1 - x0);
+    for (; x --> 0;) {
```
calc84maniac, line 142: This is cute and all but probably doesn't belong in an example. How about this equivalent (well, equivalent because `x` isn't used inside the loop):
`for(uint24_t x = (uint24_t)(x1 - x0); x > 0; x--) {`

adriweb:
looks good to us

PR #625, opened 2025-08-30
fixed useages of the hardware specific 0xE40000 address

ZERICO2005:
Removed usage of the `0xE40000` address from standard/common C functions (such as `bzero`). 

`boot_(v)snprintf` and `boot_(v)asprintf` still use the `0xE40000` address in their implementation. Which should be okay, since they are platform specific functions. 

Since `(v)snprintf` and `(v)asprintf` are standard/common C functions, I have removed the option to have them link to `boot_(v)snprintf` and `boot_(v)asprintf`.

### DIFF
--- a/src/libc/asprintf.src
+++ b/src/libc/asprintf.src
@@ -15,10 +15,14 @@ _vasprintf := __vasprintf_c
 
 else
 
+if 0
+
 _asprintf := _boot_asprintf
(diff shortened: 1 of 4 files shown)


mateoconlechuga:
Can you define the linker variable `HARDWARE=TICE` and only use `E40000` if this is defined for the hardware being used?

PR #630, opened 2025-09-06
optimized memmove

ZERICO2005:
Optimized `memmove`, as it appears to be used a lot by C++ code.

### DIFF
--- a/src/libc/memmove.src
+++ b/src/libc/memmove.src
@@ -1,0 +1,100 @@
+	assume	adl=1
+
+	section	.text
+
+	public	_memmove
+
+if PREFER_OS_LIBC
+
+_memmove := 0000A8h
+
+else
+
+if 1
+
+; Optimized for when src != dst
+_memmove:
+	; src > dst | LDIR | 32F + 15R + 1
+	; src < dst | LDDR | 35F + 12R + 2
+	; src = dst | LDDR | 35F + 12R + 2
+	; zero size |      | 24F + 12R + 2
+
+	ld	iy, -1
+	add	iy, sp
+	ld	bc, (iy + 10)
+	sbc	hl, hl
+	add	hl, bc
+	jr	nc, .zero
+	ld	hl, (iy + 7)
+	ld	de, (iy + 4)
+	sbc	hl, de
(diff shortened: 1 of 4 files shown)


src/libc/memmove.src
```diff
@@ -0,0 +1,107 @@
...
+	ld	hl, (iy + 4)
+	ret
+.copy_backwards:
+	; move HL and DE to the end
+	add	hl, de	; HL = src - 1
+	dec	de	; DE = dst - 1
+	add	hl, bc
+	ex	de, hl
+	add	hl, bc
+	ex	de, hl
```
calc84maniac, line 49: You can save 1 byte here by adding waiting to add DE to HL until after you've already added BC to it, like so:
```
	ex	de, hl
	add	hl, bc
	ex	de, hl	; HL = src - dst - 1, DE = dst + size
	add	hl, de	; HL = src + size - 1
	dec	de	; DE = dst + size - 1
```

src/libc/memmove.src
```diff
@@ -0,0 +1,107 @@
...
+	jr	nc, .zero
+	ld	hl, (iy + 7)
+	ld	de, (iy + 4)
+	sbc	hl, de
+	; src <= dst
+	jr	c, .copy_backwards
+; .copy_forwards:
+	add	hl, de
+	inc	hl
+	; src > dst or no-overlap
```
runer112, line 37: This comment and the specific documented timings for "no-overlap" suggest that all no-overlap cases end up in this code path. But the only branches that I see are for zero size and src <= dst. What am I missing?

ZERICO2005: you are correct. I have fixed those comments in the latest commit

runer112 (approved):
LGTM. But calc84 is the true optimization specialist, so maybe give him some opportunity to re-review before accepting my approval alone.

ZERICO2005:
Applied some optimizations suggested by calc84maniac to the `src == dst` optimized version. Compared with the `src != dst` version, the `src == dst` version is `1F + 1` slower when `src > dst` and `1F` slower when `src < dst`.

PR #633, opened 2025-09-13
Moved (l/i48/ll)abs/div into LIBC

ZERICO2005:
Moved the following functions from `src/crt` to `src/libc` since these functions use the C calling convention:
```
labs.src
ldiv.src
i48abs.src
i48div.src
llabs.src
lldiv.src
```

 `abs.src` and `div.src` were already inside of `src/libc`.

I also cleaned up the code of and added timings to `ineg`, `inot`, `ineg_fast`, and `inot_fast`.


ZERICO2005:
I undid the libcall absolute value functions commit to keep this PR simple. They can be added later in a different PR

mateoconlechuga:
what no these are not libc functions.

ZERICO2005:
They are the only functions in `src/crt` that follow the C calling convention to my knowledge. These functions are defined in `<stdlib.h>`.
`ldivs.src` and `lrems.src` use the LIBCALL convention, but does `ldiv.src` is not

mateoconlechuga:
Why did you remove it????

ZERICO2005:
I also figure that `__inot` and `__ineg` are likely the most commonly called routines, and that saving 2 bytes is not really worth it. Having a size optimized version would be better suited towards large and uncommon routines

adriweb:
this is just moving files and adding documentation now, so it's fine as CI passes.

Issue #634, opened 2025-09-14
broken return values in fileioc


ZERICO2005:
`ti_Rewind`, `ti_Tell`, and `ti_GetSize` state that they return `hl = -1 if failure`. However it appears that there may be typos on the `jr nz, .ret_neg_one` path of all three functions, meaning that they could return 0 instead of -1

```asm
; ti_Rewind
	ret
	scf
.ret_neg_one:
	sbc	hl, hl
	ret
; ti_Tell and ti_GetSize
	ret
.ret_neg_one:
	sbc	hl, hl
	ret
```
Found here:
https://github.com/CE-Programming/toolchain/commit/fa9a271d561efa895d3d0bbe66b7e644964a717c#r165751331
https://github.com/CE-Programming/toolchain/commit/fa9a271d561efa895d3d0bbe66b7e644964a717c#r165751346

calc84maniac:
I think 65535 wouldn't be a valid return value for `ti_Tell` because it's greater than OS_VAR_MAX_SIZE.

ZERICO2005:
> I think 65535 wouldn't be a valid return value for `ti_Tell` because it's greater than OS_VAR_MAX_SIZE.

ah gotcha. That is convenient then, so `ftell` can safely check if `ti_tell` returns 65535

ZERICO2005:
I suppose for a similar reason. `ti_GetSize` should return -1 on error instead of 0. Since 0 could be the size of a valid file but not 65535

PR #635, opened 2025-09-14
fixed return values for ti_(Rewind/Tell/GetSize) and fixed ftell

ZERICO2005:
Fixes the issues outlined in https://github.com/CE-Programming/toolchain/issues/634

`int ti_Rewind` returns `-1` or `EOF` on failure.
`ti_Tell` and `ti_GetSize` return `(uint16_t)-1` or `65535` on failure, which is larger than `OS_VAR_MAX_SIZE` or `65512`.

`ftell` calls `ti_Tell`, so I made sure that it will return `-1L` when an error occurs.


ZERICO2005:
Fixed a bug in `int ti_SetArchiveStatus(bool archive, uint8_t handle)`, which tested `archive != 0` instead of `archive & 0x1`

mateoconlechuga:
> Fixed a bug in `int ti_SetArchiveStatus(bool archive, uint8_t handle)`, which tested `archive != 0` instead of `archive & 0x1`.

Not a bug. Don't do this.

mateoconlechuga:
No. I don't freaking care what the complier does, the libraries are not "tied to the compiler". There are countless programs out there that use ICE, old ZDS complier code, etc that are not the current compiler that you just so happen to define. We are not breaking backwards compatibility for a frivolous check.

ZERICO2005:
should we change the prototype to `uint8_t`, so the compiler is aware that we are testing `archive != 0` instead of `archive & 1`?

mateoconlechuga:
Sure that seems fine

ZERICO2005:
ti_AllocString and ti_AllocEqu didn't check if the allocation routine returned NULL, which has now been fixed.

PR #637, opened 2025-09-21
implemented (u)lltod in assembly

ZERICO2005:
Implements `(u)lltod` in assembly and optimizes `(u)ltod`.

Since `long double` has 53 bits of precision, 64 bit integers may be rounded. `(u)lltod` can detect if rounding occurs and raise `FE_INEXACT`. This can be enabled/disabled with the define `__lltod_signal_FE_INEXACT := 1` at the top of `ltod.src`.

`(u)lltod` can switch between calling `__lladd_1` or having it inlined.


src/crt/ltod.src
```diff
@@ -49,54 +51,74 @@ __lltod_common:
...
+	jr	nz, __int_to_f64_shl.finish
+	inc	de
+	sbc	hl, de
+	add	hl, de
+	jr	nz, __int_to_f64_shl.finish
+	inc	bc
 	jr	__int_to_f64_shl.finish
+else
+	call	__lladd_1	; round up to even
+	jr	__int_to_f64_shl.finish
```
ZERICO2005, line 120: would it be better to call `__lladd_1` or inline it here.

PR #639, opened 2025-09-23
implemented atoi, atol, and strto(u)l(l) in assembly

ZERICO2005:
Changes:
- Implemented `atoi`, `atol`, `strtol`, `strtoul`, `strtoll`, `strtoull` in assembly.
- Optimized `llmulu_b`

```
CC per decimal digit
atoi    :  21F +  4R +  3W +  1
atol    :  32F +  4R +  3W +  1
strtol  :  79F + 13R + 12W + 21
strtoll : 103F + 10R +  9W + 33
```
`atoi`, and `atol`, now have their own dedicated routines which offer a good performance boost compared to wrapping `strtol`. This is because they don't need to track `endptr` or overflow, and they only work with base 10.

A dedicated routine for `atoll` wouldn't be much faster than `strtoll` since the main bottle neck is multiplication.

I also added `__strtoi` and `__strtoui`, which behave exactly like `strtol` and `strtoul`, but will raise ERANGE and return INT_MIN/INT_MAX and UINT_MAX respectively when the result is out of range. This is mostly useful for implementing `std::string.stoi()`, which is supposed to throw an exception if the result is out of range for `int`. We can't throw exceptions on the eZ80, but using `__strtoi` allows for `ERANGE` to be raised instead.

### DIFF
--- a/src/libc/strtoll.src
+++ b/src/libc/strtoll.src
@@ -80,6 +80,15 @@ _strtoull:
 
 	private	__strtoll_common
 
+__strtoll_common.invalid_base:
+	xor	a, a
+	ld	c, a
+	; Setting C (base) to zero ensures that cp a, c will never set carry.
+	; forcing the function to return.
+	push	af
+	; sets BC:UDE:UHL to zero
+	jr	__strtoll_common.invalid_base_hijack
+
 __strtoll_common:
 	; output: BC:UDE:UHL
 	; NC = no overflow for strtoull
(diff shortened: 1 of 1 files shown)


src/libc/strtoll.src
```diff
@@ -0,0 +1,378 @@
...
+;-------------------------------------------------------------------------------
+.other_base:
+	ld	a, (hl)		; first digit of the number
+	push	hl
+	pop	iy
+.invalid_base_hijack:
+	ld	(ix - 1), b
+	ld	bc, 0
+	ld	de, 0
+	ld	hl, 0
```
ZERICO2005, line 180: forgot to optimize this lol

ZERICO2005, line 180: forgot to optimize this lol

adriweb:
no complaints and tests pass, let's goooo

PR #648, opened 2025-10-04
optimize libload

ZERICO2005:
Applied some optimizations to libload to reduce size.

### DIFF
--- a/src/libload/libload.asm
+++ b/src/libload/libload.asm
@@ -145,9 +145,7 @@ end macro
 relocate helpers, buf + 900
 call_relative:
 	pop	ix
-	lea	ix, ix + 3
-	push	ix
-	lea	ix, ix - 3
+	pea	ix + 3
 	push	de
 	ld	de, (ix)
 	add	ix, de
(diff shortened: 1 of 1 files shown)


src/libload/libload.asm
```diff
@@ -143,16 +142,16 @@ macro relocate? name, address*
 end macro
 
 relocate helpers, buf + 900
+if 0
```
runer112: ?

src/libload/libload.asm
```diff
@@ -161,14 +160,14 @@ jump_relative:
...
-	pop	ix
-	lea	ix, ix + 3
-	jp	(ix)
+	pop	hl
+	ld	de, (hl)
+	inc	hl
+	inc	hl
+	inc	hl
+	push	hl
+	add	hl, de			; add hl, relative - 3
```
runer112: I'm a bit OOTL, how is libload (and this function) linked nowadays? If it's dynamic, I don't think we can make this change. If it's static, I guess this is fine.

src/libload/libload.asm
ZERICO2005: `rcall` is not used anywhere in `libload`. So it has been removed

src/libload/libload.asm
runer112: That sounds right, but can someone less OOTL than I am confirm? Maybe @mateoconlechuga or @calc84maniac?

PR #651, opened 2025-10-05
Implemented strlcat

ZERICO2005:
Implements POSIX `strlcat`


adriweb:
nobody complained and all new tests look good and pass, so merging I guess :P

PR #675, opened 2025-12-08
changed JP to JR in gfx_Deprecated by reordering blocks of code (saving 8 bytes)

ZERICO2005:
I was able to save 8 bytes in `gfx_Deprecated` by reordering unconditional jumps allowing everything to be in `JR` range.

I think it should be easy to prove that the code should work the exact same.


runer112 (approved):
LGTM

PR #681, opened 2026-01-02
[binutils] add binutils support

mateoconlechuga:
plz check if there's anything immediately wrong


docs/static/makefile-options.rst
```diff
@@ -115,25 +115,35 @@ These flags are passed directly to the fasmg linker.
 
     EXTRA_LDFLAGS =
 
-.. rubric:: PREFER_OS_CRT
+.. rubric:: PREFER_CE_CRT
```
calc84maniac: Is there any fallback for when the old variable name is used in existing makefiles?

src/crt/and.src
```diff
@@ -1,43 +1,27 @@
...
 
-	section	.text
-	public	__land
+	.section	.text
+	.global	__land
+	.section	.text
+	.global	__iand
 
-if PREFER_OS_CRT
+.ifdef PERFER_CE_CRT
```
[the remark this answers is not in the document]
mateoconlechuga: yeah it is lol

docs/static/makefile-options.rst
mateoconlechuga: now there is

PR #682, opened 2026-01-05
Opt memcpy memmove

ZERICO2005:
makes `memcpy` 1F faster and `memmove` 2F faster in the common case.

The timings can be adjusted slightly for `memmove` (`src != dst` version), depending on which is faster/better on average
```avrasm
    ; src >= dst | LDIR | 30F + 15R + 1 ; jr false
    ; src <  dst | LDDR | 34F + 12R + 2 ; jr true

    ; src >= dst | LDIR | 31F + 15R + 2 ; jr true
    ; src <  dst | LDDR | 33F + 12R + 1 ; jr false
```

### DIFF
--- a/src/libc/memcpy.src
+++ b/src/libc/memcpy.src
@@ -5,17 +5,13 @@
 	.global	_memcpy
 	.type	_memcpy, @function
 
-.ifdef PREFER_OS_LIBC
-
-	.set	_memcpy, 0x0000A4
-
-.else
+	; Note: TiOS memcpy works fine, but our implementation is faster
+	; .set	_memcpy, 0x0000A4
 
 _memcpy:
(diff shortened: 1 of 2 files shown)


mateoconlechuga:
Looks good, let's remove the `PREFER_OS_LIBC` option and just use these directly. I assume these optimizations are valid @calc84maniac?

src/libc/memcpy.src
```diff
@@ -12,17 +12,30 @@
 .else
 
 _memcpy:
-	ld	iy, -1
+	; size >  0     : 25F + 15R + 1 + LDIR
+	; size >= 65536 : 32F + 16R + 3 + LDIR
```
calc84maniac: This comment is no longer applicable, right?

PR #684, opened 2026-01-06
Added non-OS libc functions

ZERICO2005:
Adds implementations for these functions:
- `memchr`
- `memcmp`
- `strcat`
- `strchr`
- `strcpy` (I think its possible to optimize this further)
- `strncpy` (based off `stpncpy`, so probably could be optimized further)
- `strstr` (wraps `memmem`, so probably could be optimized further)
- `strncat`
- `strtok_r`
- `strtok` (wraps `strtok_r`)


ZERICO2005:
I'd say `longjump` and `setjump` should go into their own PR

PR #690, opened 2026-01-31
opt libload (inline ti.CpHLDE)

ZERICO2005:
Proof for the `push af \ add hl, 15 \ pop af` stuff:
https://github.com/CE-Programming/toolchain/pull/648#discussion_r2423076728
```asm
        ; Original
        ld      de, Y
        ld      hl, X
        or      a, a
        sbc     hl, de  ; HL = X -  Y, DE = Y
        add     hl, de  ; HL = X     , DE = Y

        ex      de, hl  ; HL = Y     , DE = X
        ld      de, 15  ; HL = Y     , DE =     15
        add     hl, de  ; HL = Y + 15, DE =     15
        ex      de, hl  ; HL =     15, DE = Y + 15
        pop     hl      ; HL =      ?, DE = Y + 15

        ; Optimized
        ld      de, Y
        ld      hl, X
        or      a, a
        sbc     hl, de  ; HL = X -  Y, DE = Y

        ld      hl, 15  ; HL =     15, DE = Y
        add     hl, de  ; HL = Y + 15, DE = Y
        ex      de, hl  ; HL =      Y, DE = Y + 15
        pop     hl      ; HL =      ?, DE = Y + 15
```

### DIFF
--- a/src/libload/libload.asm
+++ b/src/libload/libload.asm
@@ -235,13 +235,12 @@ check_already_loaded:
 .no_match:
 	pop	de
 	ld	hl, (end_arc_lib_locs)
-	call	ti.CpHLDE		; have we reached the end of the table?
-	push	af
-	ex	de, hl
-	ld	de, 15			; size of search entry (9=name, 3=ram ptr, 3=arc vec ptr)
+	or	a, a
+	sbc	hl, de			; have we reached the end of the table?
+
+	ld	hl, 15			; size of search entry (9=name, 3=ram ptr, 3=arc vec ptr)
 	add	hl, de
 	ex	de, hl			; end of the extraction table?
-	pop	af
 	pop	hl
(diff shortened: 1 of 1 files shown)


src/libload/libload.asm
ZERICO2005: Here `push af \ pop af` was removed. `call ti.CpHLDE` was only used to check if `HL == DE`, so we only need to preserve the Z flag for the `jr nz, .loop` branch:
- When `jr nz, .loop` is taken: All flags are destroyed by `cp a, (hl)`.
- When `jr nz, .loop` is not taken: `move_string_to_end` (which uses `cpir`) destroys Z, P, and V. The Carry flag is unused and destroyed by `call ti.ChkFindSym`

Therefore, it is safe to remove `push af \ pop af` as we only need the Z flag for `jr nz, .loop`

PR #706, opened 2026-02-11
size optimized libload start

ZERICO2005:
Saves 14 bytes by removing the LIB_FLAGS and IX restoring code for the `return to execution if there are no libs` case.

### DIFF
--- a/src/libload/libload.asm
+++ b/src/libload/libload.asm
@@ -90,10 +90,6 @@ disable_relocations
 	call	ti.MemClear
 	pop	bc
 
-	ld	a, (iy + LIB_FLAGS)
-	ld	(flag_save), a
-	ld	(ix_save), ix		; save IX since older ICE programs don't
-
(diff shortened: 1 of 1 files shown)


src/libload/libload.asm
```diff
@@ -117,19 +113,11 @@ disable_relocations
...
 	cp	a, REQ_LIB_MARKER
 	jr	z, start
-	set	optional, (iy + LIB_FLAGS)
 	cp	a, OPT_LIB_MARKER
-	jr	z, start
-	ld	a, (flag_save)
-	ld	(iy + LIB_FLAGS), a	; restore flag bits
-	ld	ix, (ix_save)		; restore IX register
+	jr	z, start_optional
 	jp	(hl)			; return to execution if there are no libs
```
ZERICO2005: both `ix` and `(iy + LIB_FLAGS)` are unmodified here, so we do not need to restore them

PR #712, opened 2026-02-25
add ez80sf to CRT

ZERICO2005:
Adds `ez80sf` functions to CRT. The compiler does not emit any `__fp*` or `__*fp` functions yet (apart from `ftoll` and `(u)lltof` aliasing `fptoll` and `(u)lltofp`).

Notes:
- `ftoll` and `(u)lltof` now use `ez80sf` routines (which are smaller and faster than the C routines used previously).
- The intention is that the compiler will emit `__f*`/`__*f` for ti/zilog floats, and `__fp*`/`__*fp` for `ez80sf` floats.
- Currently `dtof`/`ftod` alias to `dtofp`/`fptod`.
- `dcmpo` and `dcmpu` have been implemented (in C for now)
- autotests for `fcmp`, `fcmpo`, `fcmpu`, `dcmp`, `dcmpo`, `dcmpu` have been added.


ZERICO2005:
I noticed that `src/libc/fmodf.src` is implemented incorrectly so I will need to write more tests so it gets caught.

PR #738, opened 2026-03-08
Reduced libload string footprint

ZERICO2005:
I noticed that `rload`/`ld_relative` are only used with a positive offset less than 256, so I was able to save space by storing the offset in 1 byte instead of 3 bytes.

I also changed the following error messages:
`ERROR: Missing Library` --> `ERROR: Library Missing`
`ERROR: Invalid Library` --> `ERROR: Library Invalid`
This was done so all three error strings start with the same prefix of `ERROR: Library `.


runer112:
> `call` needs an absolute address so it won't work in libload

Oh, duh.

EDIT: Also it wouldn't even be helpful in this case because you want the string pointer in a register here, not on the stack. Don't know what I was thinking.

PR #748, opened 2026-03-14
Optimize `strerror` using 5-bit encoding of error strings

runer112:
- Size: 744 -> 490 bytes.
- Worst-case speed: 0.1 -> 3.2 ms.


src/libc/perror.src
```diff
@@ -0,0 +1,58 @@
...
+.L.errchar_puts:
+	ld	a, (hl)
+	or	a, a
+	ret	z
+	inc	hl
+	jr	.L.errchar_putchar
+
+	.extern	_errno
+	.extern	_errchar
+
```
ZERICO2005: I think you would also need `.extern _strerror.hijack_hl` here

PR #755, opened 2026-03-16
added equzero (same as cmpzero but only outputs Zero flag)

ZERICO2005:
`__*equzero` is identical to `__*cmpzero` except that it only outputs to the Zero flag. Currently it just aliases `__*cmpzero`, but the intent is that it could maybe link to a smaller/faster routine.

```asm
__i48cmpzero:
; CC: 13 bytes
; Minimum  :  8F + 3R + 2
; arg == 0 : 11F + 3R + 3
; Maximum  : 14F + 3R + 3
	ex	de, hl
	add	hl, de
	or	a, a
	sbc	hl, de
	ex	de, hl
	ret	nz
	sbc	hl, de
	ret	p
	inc	e
	dec	de
	ret

; implementation 1
__i48equzero:
; CC: 9 bytes
; Minimum  :  6F + 3R + 2
; arg == 0 : 10F + 3R + 2
; Maximum  : 10F + 3R + 2
	add	hl, de
	or	a, a
	sbc	hl, de
	ret	nz
	sbc	hl, de
	add	hl, de
	ret

; implementation 2
__i48equzero:
; CC: 10 bytes
; Minimum  :  6F + 3R + 2
; arg == 0 : 11F + 3R + 2
; Maximum  : 11F + 3R + 2
	or	a, a
	sbc	hl, de
	add	hl, de
	ret	nz
	sbc	hl, hl
	adc	hl, de
	ret
```

### DIFF
--- a/src/crt/dtof.src
+++ b/src/crt/dtof.src
@@ -173,7 +173,7 @@ __dtof:
 	; BC:UDE:UHL = 1 << shift
 	; (SP) = X
 	call	__lland
-	call	__llcmpzero
+	call	__llequzero
 	pop	hl
 	; DE and BC are swapped here
 	pop	bc
@@ -215,6 +215,6 @@ __dtof:
 	jr	.L.ret_copysign
 
 	.extern	__lland
-	.extern	__llcmpzero
+	.extern	__llequzero
 	.extern	__llshru
 	.extern	__lshru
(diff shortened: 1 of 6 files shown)


ZERICO2005:
yeah previously this commit also did have `__*cmpequ` as well with the same sort of idea of being `__*cmpu` but only outputting to the Z flag. But that is a much more niche optimization and likely not worth implementing into the compiler.

PR #756, opened 2026-03-16
Added `__*shift_nz` routines (for shifting by non-zero or a constant amount)

ZERICO2005:
Similar in spirit to https://github.com/CE-Programming/toolchain/pull/755.
`__*shift_nz` may allow for a small speed optimization by optionally skipping a test for a shift-by-zero.

I know that Clang/LLVM has some functionality to detect if the shift amount is non-zero. So the compiler could be able to output `__*shift_nz` when applicable. `__*shift_nz` will be emitted either when the shift amount is constant, or the shift amount is a variable that is proven to not be zero.

Calling `__*shift_nz` with a shift amount of zero is undefined behavior.

Additionally, it is always safe to convert `__*shift_nz` back to `__*shift`.

Pros:
- Small speed optimization
- If you see `call __*shift` in the compiler output, then you can be almost certain that the shift amount is by a variable instead of a constant
- If only `__lshl_nz` is used, then it might be possible to not link `__lshl` which would save 3 bytes (although this would need FASMG `require` to implement)

Cons:
- Potential for human error in hand written assembly code.

***

Here, 3F + 1 is saved by skipping the check for a shift-by-zero in `__lshl`.
```asm
__lshl:
	inc	l
	dec	l
	ret	z	; shift by zero
__lshl_nz:
	push	bc
	ld	b, l
	ex	(sp), hl
.L.loop:
	add	hl, hl
	rla
	djnz	.L.loop
	ex	(sp), hl
	pop	bc
	ret
```
Note that `__*shift_nz` aliases `__*shift` if no optimizations are possible.
```asm
__llshru_nz:
__llshru:
; Suboptimal for large shift amounts
	push	af
	push	iy
	ld	iy, 0
	add	iy, sp
	ld	a, (iy + 9)
	or	a, a
	jr	z, .L.finish	; shift by zero
; we cannot place __llshru_nz: here
	push	de
	push	hl
```

Here is a list of routines where `__*shift_nz` is faster:
- `__bshl`
- `__bshru`
- `__bshrs`
- `__lshl`
- `__lshru`
- `__lshrs`
- `__i48shru`
- `__i48shrs`

### DIFF
--- a/src/crt/dtof.src
+++ b/src/crt/dtof.src
@@ -32,7 +32,7 @@ __dtof_helper:
 	push	de
 	pop	bc
 	ld	l, 5
-	call	__lshru
+	call	__lshru_nz
 	push	bc
 	pop	hl
 .L.finish_inf_nan:
@@ -101,7 +101,8 @@ __dtof:
 	and	a, $1F
 .L.round_up:
 .L.round_down:
-	call	__llshru
+	; BC:UDE:UHL >>= 29
+	call	__llshru_nz
 	; B, C, and UDE are zero here
 	or	a, a
 	jr	z, .L.no_round
@@ -192,7 +193,7 @@ __dtof:
 	and	a, $0F
 	or	a, $10
 
-	call	__lshru
+	call	__lshru_nz
 	xor	a, a	; subnormal exponent
 	; HL = BC >> 1
 	scf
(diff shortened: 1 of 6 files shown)


ZERICO2005:
There are more important compiler issues to work on for the moment.

PR #762, opened 2026-03-23
Fixup GAS/binutils assembly

ZERICO2005:
Bug fixes:
- added missing `@function` from `.global`/`.weak`, this is necessary for preventing these warnings from `binutils`:
```
toolchain/CEdev/binutils/bin/z80-none-elf-ld: warning: strcmp.o: missing .note.GNU-stack section implies executable stack
toolchain/CEdev/binutils/bin/z80-none-elf-ld: NOTE: This behaviour is deprecated and will be removed in a future version of the linker
 ```
- removed `@function` from `.local`
- fixed `extern` --> `.extern`
- fixed `section` --> `.section`

Formatting:
- canonicalized the order of `@function` declarations (so `.type _func, @function` is always after its corresponding `.global/.weak _func`.
- Enforce using `0x012345`/`$012345` instead of `0012345h`
- fixed `.assumes` to `assumes` in comments
- switched `ADL=1`, `adl = 1`, and etc to `adl=1` so everything is consistent toolchain wide.

### DIFF
--- a/src/crt/crt0.S
+++ b/src/crt/crt0.S
@@ -66,6 +66,7 @@ __header:
 	.equ	ti.ClrTxtShd, 0x020818
 
 	.global	__start
+	.type	__start, @function
 __start:
 	ld	iy, ti.flags
 	ld	sp, (ti.onSP)			; reset stacks
@@ -100,6 +101,7 @@ __start:
 ___app_start:
 #else
 	.global	__start
+	.type	__start, @function
 __start:
 	ld	iy, ti.flags
 #endif
@@ -321,6 +323,7 @@ ___libload_libs_ret:
 #endif
 	.equ	__start._main, $ - 3
 	.global	___exithl
+	.type	___exithl, @function
 ___exithl:
 #if HAS_ATEXIT || HAS_FINI_ARRAY || HAS_ABORT
 	push	hl

### CODE REVIEW

File: src/libc/memmem.src [hunk 1]
```diff
@@ -5,7 +5,6 @@
 	.global	_memmem
 	.type	_memmem, @function
 	.local	_memcmp_fast
-	.type	_memcmp_fast, @function
```
ZERICO2005: `_memcmp_fast` is a private function, so I don't think it needs `@function`, and I don't see it used with other `.local` functions

src/ce/lz4.src
```diff
@@ -45,7 +45,6 @@ _lz4_Decompress:
 	; Fallthrough
 
 	.local	lz4_decompress_block_internal
-	.type	lz4_decompress_block_internal, @function
```
ZERICO2005: @calc84maniac is this `.local` function supposed to have `@function`?

src/libc/ceilf.src
```diff
@@ -2,8 +2,8 @@
 
 	.section	.text
 	.global	_ceil
-	.global	_ceilf
 	.type	_ceil, @function
+	.global	_ceilf
 	.type	_ceilf, @function
```
ZERICO2005: this is what I mean by canonicalizing the order of `@function`. And this makes it consistent with how it has been done elsewhere.

PR #771, opened 2026-03-30
optimized fneg

ZERICO2005:
I have optimized `fneg`, making the common case 3F faster. The main decision left is for inputs `[+0.0f, +FLT_MIN * 2.0f)`, since we can make the case of `input == 0.0f` slightly faster (-1F + 1) at the cost of `[+FLT_TRUE_MIN, +FLT_MIN * 2.0f)` being slightly slower (+1F + 1)

Previous code:
```
    ; 13 bytes
    ; A != 0        :  8F + 3R + 0W + 2
    ; A:UBC == 0.0f : 11F + 6R + 3W + 2
    ; A:UBC != 0.0f : 14F + 6R + 3W + 2
```
Option 1 (optimize equally):
```
    ; 12 bytes
    ; A != 0        :  5F + 3R + 0W + 2
    ; A:UBC == 0.0f : 13F + 6R + 3W + 3
    ; A:UBC != 0.0f : 13F + 6R + 3W + 3
```
Option 2 (optimize input == 0.0f):
```
    ; 13 bytes
    ; A != 0        :  5F + 3R + 0W + 2
    ; A:UBC == 0.0f : 12F + 6R + 3W + 3
    ; A:UBC != 0.0f : 14F + 6R + 3W + 3
```

### DIFF
--- a/src/crt/fneg.src
+++ b/src/crt/fneg.src
@@ -23,13 +23,23 @@
 	.type	__fneg, @function
 
 __fneg:
-	or	a,a
-	jr	nz,__noexit
+	; Common case (A != 0) :  5F + 3R + 0W + 2
+	; A:UBC == 0.0f        : 12F + 6R + 3W + 3
+	; positive subnormal   : 14F + 6R + 3W + 3
+	add	a, $7F
+	inc	a
+	ret	po
+	; input == [+0.0f, +FLT_MIN * 2.0f)
+	; we need to make sure we do not output -0.0f, so we need to make sure that the mantissa is not zero.
+	; A is $80 here
+	rla			; set carry (and set A to $00)
 	push	hl
-	sbc	hl,hl
-	sbc	hl,bc
+	sbc	hl, hl
+	add	hl, bc
 	pop	hl
-	ret	z
-__noexit:
-	xor	a,80h
+.if 1
+	; we can optionally return here to make the A:UBC == 0.0f case slightly faster.
+	ret	nc
+.endif
+	rra			; A = (BC != 0) ? $80 : $00
 	ret


src/crt/fneg.src
```diff
@@ -23,13 +23,23 @@
...
-	or	a,a
-	jr	nz,__noexit
+	; Common case (A != 0) :  5F + 3R + 0W + 2
+	; A:UBC == 0.0f        : 12F + 6R + 3W + 3
+	; positive subnormal   : 14F + 6R + 3W + 3
+	add	a, $7F
+	inc	a
+	ret	po
+	; input == [+0.0f, +FLT_MIN * 2.0f)
+	; we need to make sure we do not output -0.0f, so we need to make sure that the mantissa is not zero.
```
runer112, line 33: Why?

ZERICO2005: Zilog floats do not support signed zeros, otherwise `__fneg` would simply be `xor a, $80`.

runer112: Maybe this PR isn't the right place to attempt this, but what would it take to support signed zeros?

ZERICO2005: https://github.com/CE-Programming/toolchain/pull/712
ez80sf worked on by Jacobly and Calc84Maniac adds proper ieee 754 support. Including signed zeros, and rounding to nearest with ties to even (instead of ties away from zero).

Not to mention, some `libc` functions return signed zeros (like `truncf.src` and others do not `ldexpf.src`, but nothing has broken yet so `¯\_(ツ)_/¯`

ZERICO2005: but yeah, this PR for `__fneg` does make negation for Zilog floats faster for the average case (`8F + 3R + 2` --> `5F + 3R + 2`).

runer112 (approved):

PR #777, opened 2026-04-10
optimize `__lshr(u/s)`

ZERICO2005:
- combined both lshrs and lshru into one file
- Renamed: `ishrs.src`, `lshrs.src`, and `llshrs.src` to `ishr.src`, `lshr.src`, and `llshr.src` since they contain both the signed and unsigned shift routines.
- Optimized `lshrs` and `lshru`. The worst case is 61.8% faster now (assuming F = 3, R = 3, W = 2)

```
; previous CC for __lshrs:
; L == 0:   4F +  3R +  0W +   2
; L == 1:  28F + 12R +  9W +   5
; L >= 2:  28F + 12R +  9W +   5 + (15F + 3R + 3W + 4) * (L - 1)
; L == 2:  43F + 15R + 12W +   9
; Max CC: 478F + 43R + 40W + 125

; optimized CC for __lshrs:
; L == 0:   4F +  3R +  0W +  2
; L == 1:  33F + 13R + 10W +  2
; L >= 2:  33F + 13R + 10W +  2 + (10F + 1) * (L - 1)
; L == 2:  44F + 13R + 10W +  3
; Max CC: 334F + 13R + 10W + 32
```


src/crt/lshr.src
```diff
@@ -0,0 +1,85 @@
...
+	rr	c	; C
+	djnz	.L.loop
+	ld	(hl), a
+	ld	l, c
+	pop	bc
+	ld	b, d
+	ld	c, l
+	ld	a, e
+	pop	de
+	pop	hl
```
runer112: \- 7F + 2R + 2W

```suggestion
	push	de
	ld	e, a
	ld	a, c
	push	bc
	ld	b, l
	dec	sp
	ex	(sp), hl
.L.loop:
	sra	e	; A
	.local	__lshrs.hijack_lshru
__lshrs.hijack_lshru:
	rr	h	; UBC
	rr	l	; B
	rra		; C
	djnz	.L.loop
	ex	(sp), hl
	inc	sp
	pop	bc
	ld	c, a
	ld	a, e
	pop	de
```

src/crt/lshr.src
```diff
@@ -0,0 +1,85 @@
...
+	push	de
+	push	bc
+	ld	d, b
+	ld	b, l
+	ld	hl, 2
+	add	hl, sp		; clears carry
+	rra			; cheaper than doing srl e later on
+	ld	e, a
+	ld	a, (hl)
+	; srl	e
```
runer112: \- 6F + 2R + 2W

```suggestion
	push	de
	ld	e, a
	ld	a, c
	push	bc
	ld	b, l
	dec	sp
	ex	(sp), hl
	srl	e	; A
```

ZERICO2005: Should be `inc sp \ ex (sp), hl` then `ex (sp), hl \ dec sp` I think

runer112: Err yeah, I think you're right about that. Sorry, my z80 is rusty.

ZERICO2005: That does also swap UHL with the preserved E on the stack, but as long as the value of UHL isn't modified in the loop everything should be okay

src/crt/lshr.src
```diff
@@ -0,0 +1,85 @@
...
+
+	.set	__lshrs, 0x0001E8
+
+.else
+
+__lshrs:
+; (int32_t)A:UBC >>= L
+; Suboptimal for large shift amounts
+; CC:
+; L == 0:   4F +  3R +  0W +  1
```
ZERICO2005: it was supposed to be `4F + 3R + 2`. Silly me :)

src/crt/lshr.src
ZERICO2005: implemented!

src/crt/lshr.src
ZERICO2005: implemented!

src/crt/lshr.src
ZERICO2005: typo fixed :)

runer112 (approved):

PR #790, opened 2026-04-17
Fix #789 and add constructor/destructor/init_priority tests

### LINKED ISSUE #789: `__attribute__((constructor(N)))` does not set `HAS_INIT_ARRAY = 1`
```c
#include <ti/screen.h>
#include <ti/getcsc.h>

__attribute__((constructor(123)))
void ctor_func(void) {
    os_ClrHome();
    os_PutStrFull(__FUNCTION__);
    while (!os_GetCSC());
}

__attribute__((destructor))
void dtor_func(void) {
    os_ClrHome();
    os_PutStrFull(__FUNCTION__);
    while (!os_GetCSC());
}


int main(void) {
    os_ClrHome();
    os_PutStrFull("main_func");
    while (!os_GetCSC());

    return 0;
}
```
In this program, only `main`, and `dtor_func` run. This is because `cedev-obj` effectively only detects `__attribute__((constructor))` and `__attribute__((destructor))`, but not `__attribute__((constructor(N)))` or `__attribute__((destructor(N)))`

ZERICO2005:
Fixes https://github.com/CE-Programming/toolchain/issues/789

`cedev-obj` currently rejects `NULL` strings, so I also made it also reject empty `""` strings (especially since `strncmp(str, "", len)` always returns 0, thereby matching any string)

Added test for:
```
__attribute__((constructor))
__attribute__((constructor(N)))

__attribute__((destructor))
__attribute__((destructor(N)))

__attribute__((init_priority(N)))
```

### DIFF
--- a/tools/cedev-obj/src/elf_inspect.c
+++ b/tools/cedev-obj/src/elf_inspect.c
@@ -464,6 +464,30 @@ bool elf_has_section(struct elf_file *elf, const char *section_name)
     return false;
 }
 
+bool elf_has_section_prefix(struct elf_file *elf, const char *section_prefix)
+{
+    if (!elf || !section_prefix)
+    {
+        return false;
+    }
+
+    const size_t prefix_len = strlen(section_prefix);
+    for (uint16_t i = 0; i < elf->ehdr.e_shnum; i++)
+    {
+        if (elf->section_headers[i].sh_name < elf->shstrtab_size)
+        {
+            const char *name = elf->shstrtab + elf->section_headers[i].sh_name;
+            if (strncmp(name, section_prefix, prefix_len) == 0)
+            {
+                /* Section exists - check if it's non-empty */
+                return elf->section_headers[i].sh_size > 0;
+            }
+        }
+    }
+
+    return false;
+}
+
 const char *elf_get_error(struct elf_file *elf)
 {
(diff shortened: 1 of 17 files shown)


tools/cedev-obj/src/elf_inspect.c
```diff
@@ -464,6 +479,30 @@ bool elf_has_section(struct elf_file *elf, const char *section_name)
...
+            if (strncmp(name, section_prefix, prefix_len) == 0)
+            {
+                /* Section exists - check if it's non-empty */
+                return elf->section_headers[i].sh_size > 0;
+            }
+        }
+    }
+
+    return false;
+}
```
runer112, line 504: If you wanted to, you could implement both of these with a single shared function that uses `strncmp` and accepts the length as a parameter.

runer112 (approved):

tools/cedev-obj/src/elf_inspect.c
ZERICO2005: I think this function may have to be changed to search for exactly "foobar" or a "foobar." prefix instead of for a "foobar" prefix

mateoconlechuga (approved):

mateoconlechuga (approved):

PR #791, opened 2026-04-18
[fatdrvce] converted division by a power of two into a bitshift

### LINKED ISSUE #757: division/remainder can be replaced with bitshifts in `fatdrvce`
I noticed that division is used instead of bit shifts for getting the quotient/remainder by a power of 2.
`util_cluster_entry_to_block` does 32bit unsigned division by 128
`util_ceil_byte_size_to_block_size` does 32bit unsigned division and remainder by 512

I think it would be beneficial to rewrite these functions to use bit shifts as division is a very slow operation to perform.

ZERICO2005:
TL:DR turns 32bit division into a bitshift

Resolves https://github.com/CE-Programming/toolchain/issues/757

### DIFF
--- a/src/fatdrvce/fatdrvce.asm
+++ b/src/fatdrvce/fatdrvce.asm
@@ -2495,32 +2495,49 @@ util_get_cluster_offset:
 
 ;-------------------------------------------------------------------------------
 util_cluster_entry_to_block:
-	ld	e,a
-	xor	a,a
-	ld	bc,128
-	call	ti._ldivu
-	ld	bc,(yfat.fat_pos)
+	; input:
+	; - A:UHL
+	; output:
+	; - A:UHL = floor(A:UHL / 128) + (yfat.fat_pos)
+	; destroys:
+	; - BC, flags
+	push	hl
+	pop	bc
+	ld	l,7
+	call	ti._lshru
+	ld	hl,(yfat.fat_pos)
 	add	hl,bc
-	adc	a,e
+	adc	a,0
 	ret
 
 ;-------------------------------------------------------------------------------
 util_ceil_byte_size_to_block_size:
+	; input:
+	; - A:UHL
+	; output:
+	; - A:UHL = ceil(A:UHL / 512)
+	; - A = 0
+	; destroys:
+	; - BC, flags
 	compare_auhl_zero
 	ret	z
-	ld	e,a
-	push	hl,de
-	xor	a,a
-	ld	bc,512
-	push	bc
-	call	ti._lremu
-	compare_hl_zero
-	pop	bc,de,hl
+	; test if the low 9 bits are non-zero
+	inc	l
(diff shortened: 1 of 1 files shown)


src/fatdrvce/fatdrvce.asm
```diff
@@ -2495,32 +2495,49 @@ util_get_cluster_offset:
...
+	pop	bc
+	ld	l,9
+	call	ti._lshru
+	push	bc
+	pop	hl
 	pop	af
+	ld	a,0
 	ret	z
+	; round up
 	inc	hl
```
runer112:
```asm
	; if A:UHL == 0, abort
	compare_hl_zero
	jr	nz, .dec_hl
	or	a, a
	ret	z

	; A:UHL--
	dec	a
.dec_hl:
	dec	hl

	; A:UHL /= 512
	push	hl
	pop	bc
	ld	l, 9
	call	ti._lshru
	push	bc
	pop	hl

	; A:UHL++
	inc	hl

	ret
```

ZERICO2005: Good point. That trick works since we have already filtered out the A:UHL == 0 case

src/fatdrvce/fatdrvce.asm
```diff
@@ -2495,32 +2495,50 @@ util_get_cluster_offset:
...
+	; A:UHL /= 512
+	push	hl
+	pop	bc
+	ld	l,9
+	call	ti._lshru
+	push	bc
+	pop	hl
+
+	; A:UHL++
 	inc	hl
```
calc84maniac, line 2541: It turns out that it's possible to inline both shifts while taking the same amount of space as the current implementation:
```suggestion

	; shift A:UHL left by 1 and shift the 33-bit result right by 8
	add	hl,hl
	rla
	push	af
	inc	sp
	push	hl
	inc	sp
	pop	hl
	inc	sp
	ccf
	sbc	a,a
	; the increment to set A to 0 or 1 is deferred until the adc below
	ld	bc,(yfat.fat_pos)
	add	hl,bc
	adc	a,1
	ret

;-------------------------------------------------------------------------------
util_ceil_byte_size_to_block_size:
	; input:
	; - A:UHL
	; output:
	; - A:UHL = ceil(A:UHL / 512)
	; - A = 0
	; destroys:
	; - BC, flags

	; add 511 to A:UHL and shift the 33-bit result right by 9
	ld	bc,511
	add	hl,bc
	adc	a,0
	rra
	push	af
	inc	sp
	push	hl
	inc	sp
	pop	hl
	inc	sp
	rr	h
	rr	l
	xor	a,a
```

ZERICO2005:
Unrelated, but I noticed that `ret` is unreachable here?
```
	adc	a,b
	jq	.traversefat
	ret
.unallocatedcluster:
	ld	a,128
```

PR #792, opened 2026-04-20
Optimize crt0

ZERICO2005:
Size optimizes the timer initialization/clearing code, and optimizes the atexit function loop.

Also formatted unconditional branches.

One important change was that I swapped the order of when the `atexit`/`on_exit` function and `free` get called. So now `free` is called after the `atexit`/`on_exit` function (which probably is more technically correct anyways?)

### DIFF
--- a/src/crt/crt0.S
+++ b/src/crt/crt0.S
@@ -291,12 +291,11 @@ ___libload_libs_ret:
 
 	; Set the counter to zero
 	.errif	(ti.mpTmr1Counter & 0xFFFF) != 0
-	ld	l, h				; ld hl, ti.mpTmr1Counter
-	ld	b, 4
+	ld	l, (ti.mpTmr1Counter + 4) & 0xFF
 .L.zeroize_counter:
+	dec	l
 	ld	(hl), h
-	inc	hl
-	djnz	.L.zeroize_counter
+	jr	nz, .L.zeroize_counter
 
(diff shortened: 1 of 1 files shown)


src/crt/crt0.S
```diff
@@ -343,38 +351,44 @@ _exit:
...
-	call	_free
-	pop	bc				; reset SP
-	pop	hl				; func
+	ld	hl, (ix + 1 * 3)		; func
 	; atexit  : void (*func)(void)
 	; on_exit : void (*func)(int status, void *arg)
 	call	__indcallhl
+	ex	(sp), ix
+	call	_free
 	pop	bc				; reset SP
```
ZERICO2005, line 365: swapped the order of `_free` and `__indcallhl` here. Is this okay?

mateoconlechuga: I'm more concerned that this doesn't use the custom allocator if it is defined, I knew there was something wrong with that previous change.

ZERICO2005: I think this code does use the custom free. I made my custom free print a message when it was called in test/standalone/custom_allocator

ZERICO2005: I also made sure that `crt0.S` was the only place that `free` is called in that test program

mateoconlechuga: Sounds good, this change looks fine then

mateoconlechuga (approved):

PR #801, opened 2026-04-25
Add overlap support for gfx_CopyRectangle

### LINKED ISSUE #660: Allow overlap with gfx_CopyRectangle
Currently, `gfx_CopyRectangle` does not allow for two buffers to overlap. However, it does not appear to be too difficult to allow overlapping buffers.

This basically means implementing `gfx_CopyRectangle` like `memmove` instead of `memcpy`. Copying forwards with LDIR when `src > dst`, and copying backwards with LDDR when `src < dst`

ZERICO2005:


mateoconlechuga:
This requires a library version increase

ZERICO2005:
Good point!

adriweb:
Mateo said it's fine, and tests pass, merging.

PR #807, opened 2026-06-19
Fixed <stdio.h> file and dereferencing bugs

ZERICO2005:
Fixed the following issues:
```
[P2] clearerr indexes file streams off by one: src/libc/clearerr.c:13. fopen stores slot 1 at _file_streams[0], but clearerr uses _file_streams[stream->slot]; slot 5 writes past FOPEN_MAX.
[P2] fread(stdin, ...) reports full success after EOF/partial reads: src/libc/fread.c:20. The loop stops on EOF but returns count unconditionally instead of completed elements.
[P2] fwrite(stdout, size=0, ...) divides by zero: src/libc/fwrite.c:17. For zero-sized writes, the loop does nothing, then returns num / size.
```

Also fixed:
- Dereferencing `FILE * stream` when it is `NULL`, `stdin`, `stdout`, or `stderr`.
- return value for `fseek` and `ftell`

### DIFF
--- a/src/libc/clearerr.c
+++ b/src/libc/clearerr.c
@@ -1,5 +1,7 @@
 #include "__fileioc_stdio.h"
 
+#include <stdint.h>
+
 void __attribute__((weak)) clearerr(FILE *stream)
 {
     if (stream == NULL ||
@@ -9,7 +11,7 @@ void __attribute__((weak)) clearerr(FILE *stream)
     {
         return;
     }
-
-    _file_streams[stream->slot].eof = 0;
-    _file_streams[stream->slot].err = 0;
+    uint8_t index = stream->slot - 1;
+    _file_streams[index].eof = 0;
+    _file_streams[index].err = 0;
 }
--- a/src/libc/fclose.c
+++ b/src/libc/fclose.c
@@ -1,5 +1,7 @@
 #include "__fileioc_stdio.h"
 
+#include <stdint.h>
+
 int __attribute__((weak)) fclose(FILE *stream)
 {
     ti_var_t slot;
@@ -14,8 +16,13 @@ int __attribute__((weak)) fclose(FILE *stream)
 
     slot = stream->slot;
 
-    _file_streams[slot - 1].slot = 0;
-
     int status = ti_Close(slot);
-    return (status == 0) ? EOF : 0;
+    if (status == 0)
+    {
+        // failed to close file
+        return EOF;
+    }
+    uint8_t index = slot - 1;
+    _file_streams[index].slot = 0;
+    return 0;
 }
--- a/src/libc/feof.c
+++ b/src/libc/feof.c
@@ -2,5 +2,9 @@
 
 int __attribute__((weak)) feof(FILE *stream)
 {
+    if (stream == NULL || stream == stdin || stream == stdout || stream == stderr)
+    {
+        return 0;
+    }
     return stream->eof;
 }
--- a/src/libc/fgetc.c
+++ b/src/libc/fgetc.c
@@ -2,25 +2,16 @@
 
 int __attribute__((weak)) fgetc(FILE *stream)
 {
-    int c;
-
-    if (stream == NULL)
+    if (stream == NULL || stream == stdout || stream == stderr)
     {
         return EOF;
     }
+    if (stream == stdin)
+    {
+        return getchar();
+    }
 
-    if (stream == stdout || stream == stderr)
-    {
-        c = EOF;
-    }
-    else if (stream == stdin)
-    {
-        c = getchar();
-    }
-    else
-    {
-        c = ti_GetC(stream->slot);
-    }
+    int c = ti_GetC(stream->slot);
 
     if (c == EOF)
     {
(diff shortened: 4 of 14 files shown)


src/libc/include/stdio.h
```diff
@@ -61,9 +61,9 @@ size_t fread(void *__restrict ptr, size_t size, size_t count, FILE *__restrict s
 
 size_t fwrite(const void *__restrict ptr, size_t size, size_t count, FILE *__restrict stream);
 
-long int ftell(FILE *stream) __attribute__((__warn_unused_result__));
```
ZERICO2005, line 64: `long int` is archaic, so I have replaced it with just plain `long`.

src/libc/clearerr.c
```diff
@@ -9,7 +11,7 @@ void __attribute__((weak)) clearerr(FILE *stream)
     {
         return;
     }
-
-    _file_streams[stream->slot].eof = 0;
-    _file_streams[stream->slot].err = 0;
+    uint8_t index = stream->slot - 1;
+    _file_streams[index].eof = 0;
+    _file_streams[index].err = 0;
```
ZERICO2005, line 16: fixed off by one error, since `stream->slot` is 1-indexed instead of 0-indexed

src/libc/fclose.c
```diff
@@ -14,8 +16,13 @@ int __attribute__((weak)) fclose(FILE *stream)
 
     slot = stream->slot;
 
-    _file_streams[slot - 1].slot = 0;
-
     int status = ti_Close(slot);
```
ZERICO2005, line 19: Only closing the file in `libc` if and only if `fileioc` was able to successfully close the file

src/libc/feof.c
```diff
@@ -2,5 +2,9 @@
 
 int __attribute__((weak)) feof(FILE *stream)
 {
+    if (stream == NULL || stream == stdin || stream == stdout || stream == stderr)
+    {
+        return 0;
+    }
```
ZERICO2005, line 8: We cannot dereference `stream` if it is `NULL`, `stdin`, `stdout`, or `stderr`

src/libc/fgetc.c
```diff
@@ -2,26 +2,17 @@
...
-    {
-        c = ti_GetC(stream->slot);
+        return getchar();
     }
 
+    int c = ti_GetC(stream->slot);
+
     if (c == EOF)
     {
         stream->eof = 1;
```
ZERICO2005, line 18: I fixed things so `stdin`, `stdout`, and `stderr` are not dereferenced here

src/libc/fread.c
```diff
@@ -16,20 +17,22 @@ size_t __attribute__((weak)) fread(void *ptr, size_t size, size_t count, FILE *_
         int c;
         char *p = ptr;
         size_t len = size * count;
+        size_t bytes_read = 0;
 
-        for (; len > 0; len--)
+        for (; len --> 0;)
```
ZERICO2005: This is a more common idiom for writing the decrementing count

src/libc/fseek.c
```diff
@@ -9,6 +9,5 @@ int __attribute__((weak)) fseek(FILE *stream, long int offset, int origin)
     {
         return -1;
     }
-
-    return ti_Seek((int)offset, origin, stream->slot);
+    return (ti_Seek((int)offset, origin, stream->slot) == EOF) ? -1 : 0;
```
ZERICO2005, line 12: `fseek` returns 0 on success and non-zero otherwise

src/libc/ftell.c
```diff
@@ -14,5 +14,5 @@ long int __attribute__((weak)) ftell(FILE *stream)
     // ti_Tell shouldn't return a value greater than OS_VAR_MAX_SIZE (65512) unless an error occurs
     uint16_t ret = ti_Tell(stream->slot);
     // Convert a result of UINT16_MAX to -1L, leaving other results the same
-    return ((uint16_t)ret + 1) - 1L;
```
ZERICO2005, line 17: This had an integer promotion bug. `(uint16_t)ret + 1` results in `int`, not `uint16_t`

test/standalone/asprintf_fprintf/src/main.c
```diff
@@ -566,14 +567,14 @@ int memccpy_tests(void) {
...
+        C(fprintf(file, "%s%c", names_only, '\n') > 0);
     } else {
         printf("Error Buffer is too small.\n");
     }
 
-    fseek(file, 0, SEEK_END);
+    C(fseek(file, 0, SEEK_END) == 0);
     int file_size = ftell(file);
-    fseek(file, 0, SEEK_SET);
+    C(fseek(file, 0, SEEK_SET) == 0);
```
ZERICO2005, line 577: Added some basic return value checks. Mostly important `fseek` since it returned the wrong value prior.

src/libc/fread.c
mateoconlechuga: No it fucking isn't

ZERICO2005: Reverted that change

Issue #754, opened 2026-03-16
ignore line spacing flag does not work


ZERICO2005:
details mentioned here:
https://github.com/CE-Programming/toolchain/pull/661

***

Doing a search on github for `fontlib_SetFont` https://github.com/search?q=fontlib_SetFont&type=code
I found these results:
- `fontlib_SetFont(font, 0)`: at least 25 times
- `fontlib_SetFont(font, FONTLIB_IGNORE_LINE_SPACING)`: used only once in https://github.com/the-real-mcarn/stopresisting-ti84

So it seems like most users did not use the `FONTLIB_IGNORE_LINE_SPACING` flag, but `fontlib_SetFont(font, 0)` has a 99.6% chance as being treated as `fontlib_SetFont(font, FONTLIB_IGNORE_LINE_SPACING)`

***

My proposal is to:
- Have `fontlib_SetFont` ignore line spacing with a 100% chance instead of a 99.6% chance.
- Create a new routine (`fontlib_LoadFont` or etc) that properly handles the flags.
- Discourage `fontlib_SetFont` in new code
- Rename `fontlib_SetFont` to `fontlib_Deprecated` or etc, and have `#define fontlib_SetFont fontlib_Deprecated` with a warning so developers can update their code

mateoconlechuga:
I like the proposal

ZERICO2005:
There are also some other things we should consider. This enum is used exclusively for `fontlib_SetFont`. Since the value of the enum is ignored, we can safely change the value of the `enum` to something else. 
```
typedef enum {
    /**
     * If set, then the space above/below metrics specified in the font will be
     * ignored and set to zero.  You can still manually set them later.
     */
    FONTLIB_IGNORE_LINE_SPACING = 0x01
} fontlib_load_options_t;
```

I also noticed that the current `fontlib_SetFont` code would theoretically check if the flags are non-zero (which is wasteful since that only allows one boolean of configuration), so my current `fontlib_LoadFont` code will check if a bit is set or not (so we can have support 8 or 24 flags in total instead of just 1).

PR #808, opened 2026-06-21
Add fontlib_LoadFont to replace fontlib_SetFont

ZERICO2005:
Addresses https://github.com/CE-Programming/toolchain/issues/754

- Adds `fontlib_LoadFont` which correctly handles flags when loading a font.
- Changed all occurances of `fontlib_SetFont` with `fontlib_LoadFont` in the documentation.
- Developers will get a deprecation warning when compiling code that uses `fontlib_SetFont`, and are encouraged to replace `fontlib_SetFont` with `fontlib_LoadFont`.
- `fontlib_SetFont(font, flags)` will be equivalent to calling `fontlib_LoadFont(font, FONTLIB_IGNORE_LINE_SPACING)` for backwards compatibility purposes.


src/fontlibc/fontlibc.h
```diff
@@ -162,7 +161,7 @@ typedef struct fontlib_metadata_t {
  *  unsigned char baseline;
  *  fontlib_font_t *my_font = fontlib_GetFontByStyle("FONTPACK", 12, 12,
  *      FONTLIB_NORMAL, FONTLIB_NORMAL, FONTLIB_SERIF, 0);
- *  if (!my_font || !fontlib_SetFont(my_font))
```
ZERICO2005, line 165: Was the previous prototype `fontlib_SetFont(void*)`? If so that means *really* old code will not be aware of the extra parameter

PR #811, opened 2026-07-09
crt saturate add sub

ZERICO2005:
LLVM does not emit these currently.

Adds routines for signed/unsigned add/subtract overflow/saturation (except for 8 bit since calling convention is undefined for that width)

`__*(add/sub)u_overflow` outputs to NC/C
`__*(add/sub)s_overflow` outputs to PO/PE
`__*(add/sub)u_sat` performs unsigned saturation
`__*(add/sub)s_overflow` performs signed saturation

Some of the overflow routines alias the regular `__*add`/`__*sub` routines

### DIFF
--- a/src/crt/i48add.src
+++ b/src/crt/i48add.src
@@ -6,9 +6,24 @@
 	.assume	adl=1
 
 	.section	.text
+
 	.global	__i48add
 	.type	__i48add, @function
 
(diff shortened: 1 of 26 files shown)


ZERICO2005:
ima need to fix some typoes

